PR #24, opened 2014-12-17
Add a RUN step to install freshplayerplugin - Flash Player for Firefox

echernyavskiy:


rubytester:
We use Firefox without flash. Firefox complains that flash plugin is not installed but this has not caused any functional testing issues for us.  I would recommend adding flash in a downstream container.

rubytester:
I assume there is no more tracktion on this pull request and I want to close it and defer plugin installation in a downstream container based on `node-firefox`. Opinions?

mtscout6:
That's fair

rubytester:
spoiled milk at this point, reopen or start a new conversation

PR #38, opened 2015-01-28
Add PhantomJS to list of possible nodes.

aholtventuretech:
Added NodePhantomJS, which includes Dockerfile and entry point script for running PhantomJS as a node within docker.


aholtventuretech:
Note:  When running tests on PhantomJS using selenium grid, always clear out all cookies between tests, before starting a test, or as part of test cleanup.  This is due to a known issue with PhantomJS https://github.com/detro/ghostdriver/issues/170

Example:  As part of my wrapped "close driver" method for test cleanup, I simply added this line before calling the quit() method on the driver:  getBaseDriver().manage().deleteAllCookies();

NodePhantomJS/Dockerfile
```diff
@@ -0,0 +1,41 @@
+FROM selenium/base:2.44.0
```
mtscout6: Is there a reason you chose to derive from base and not node-base? For example both the Chrome and Firefox nodes derive from node-base: https://github.com/SeleniumHQ/docker-selenium/blob/master/NodeChrome/Dockerfile#L1

mtscout6: That base image takes care of about half of this dockerfile

NodePhantomJS/Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+#===================
+ENV TZ "US/Pacific"
+RUN echo "US/Pacific" | sudo tee /etc/timezone \
+  && dpkg-reconfigure --frontend noninteractive tzdata
+
+#==============
+# PhantomJS
+#==============
+RUN apt-get update -y
+RUN apt-get install bzip2 curl libfreetype6 libfontconfig1  -y
```
mtscout6: Do you really need to install curl? `wget` is already installed in the [base image](https://github.com/SeleniumHQ/docker-selenium/blob/master/Base/Dockerfile#L19) There should be no reason to install two different tools for downloading internet content.

NodePhantomJS/entry_point.sh
```diff
@@ -0,0 +1,19 @@
...
+fi
+
+function shutdown {
+  kill -s SIGTERM $NODE_PID
+  wait $NODE_PID
+}
+
+ip="$(ifconfig | grep -A 1 'eth0' | tail -1 | cut -d ':' -f 2 | cut -d ' ' -f 1)"
+
+phantomjs --webdriver=$ip:8080 --webdriver-selenium-grid-hub=http://$HUB_PORT_4444_TCP_ADDR:$HUB_PORT_4444_TCP_PORT
```
mtscout6: If you let WebDriver launch this than you should avoid your cookie problem since the process would regularly be restarted. You should include a config.json similar to the Chrome and Firefox images to accomplish this: https://github.com/SeleniumHQ/docker-selenium/blob/master/NodeChrome/config.json

aholtventuretech: Phantomjs (as far as I know) still is not fully integrated into WebDriver.  It is, however, integrated with Ghostdriver, which allows it to register itself as a node to a selenium grid hub.

You could try telling webdriver to launch phantomjs with a config, but it wouldn't work.

https://groups.google.com/forum/#!topic/webdriver/HA5C9Jnpr7g

NodePhantomJS/Dockerfile
aholtventuretech: A large chunk of this dockerfile is similar to the node-base dockerfile, save for one difference.

I chose to derive from base rather than node-base because neither VNC nor Xvfb are needed by a phantomjs node.  It is already headless, so there would really be no graphical view or reason to VNC into that node.

NodePhantomJS/Dockerfile
aholtventuretech: Fixing now.  Will push update when complete.

LouTerrailloune:
@aholtventuretech Why do you use v1.9.8 and not 2.0 ? Lack of precompiled binaries ?

NodePhantomJS/entry_point.sh
LouTerrailloune: Maybe you can also handle the case when there is no hub defined, run it as standalone.

tedzipfworks:
I tried this pull request and it did not work for me, the container does not start up. Any one else have better luck?

PR #52, opened 2015-03-30
Adds support for -remoteHost parameter for nodes. This is required when ...

charford:
...node is running on separate host from hub, and also has multiple network interfaces. This is because selenium by default sets -remoteHost to first available network interface, which may not be the one that you want to use. In my use case, we have two network interfaces, and it's the 2nd one listed that we use for our remoteHost ip. More info about this parameter available here: http://automatictester.co.uk/2012/10/27/selenium-grid-2-0-and-remotehost-parameter/. 

To enable, add -e REMOTE_HOST="http://1.2.3.4:5555" to docker run command for node, replacing 1.2.3.4 with the accesible IP of your node.

### DIFF
--- a/NodeBase/entry_point.sh
+++ b/NodeBase/entry_point.sh
@@ -16,12 +16,19 @@ function shutdown {
   wait $NODE_PID
 }
 
+REMOTE_HOST_PARAM=""
+if [ ! -z "$REMOTE_HOST" ]; then
+  echo "REMOTE_HOST variable is set, appending -remoteHost"
+  REMOTE_HOST_PARAM="-remoteHost $REMOTE_HOST"
+fi
+
 # TODO: Look into http://www.seleniumhq.org/docs/05_selenium_rc.jsp#browser-side-logs
 
 xvfb-run --server-args="$DISPLAY -screen 0 $GEOMETRY -ac +extension RANDR" \
   java -jar /opt/selenium/selenium-server-standalone.jar \
     -role node \
     -hub http://$HUB_PORT_4444_TCP_ADDR:$HUB_PORT_4444_TCP_PORT/grid/register \
+    ${REMOTE_HOST_PARAM} \
     -nodeConfig /opt/selenium/config.json &
 NODE_PID=$!
 
--- a/NodeChromeDebug/entry_point.sh
+++ b/NodeChromeDebug/entry_point.sh
@@ -16,6 +16,12 @@ function shutdown {
   wait $NODE_PID
 }
 
+REMOTE_HOST_PARAM=""
+if [ ! -z "$REMOTE_HOST" ]; then
+  echo "REMOTE_HOST variable is set, appending -remoteHost"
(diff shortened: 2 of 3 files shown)


charford:
This also allows us to change the port that node is running on (default is 5555). Example:

docker run -d -p 5556:5555 -e REMOTE_HOST="http://1.2.3.4:5556" -e HUB_PORT_4444_TCP_ADDR="1.2.3.4" -e HUB_PORT_4444_TCP_PORT="4444" --name firefox-node selenium/node-firefox

unusorin:
ETA for this getting merged ?

sushanghimire:
@charford I have run into same issue, I have run node using same -remoteHost flag, node can register to hub and hub can check node's status. However when I run a test, Chrome crashes (unknown error: Chrome failed to start: crashed). In log I saw it is still trying to use docker's internal IP

System info: host: '72e88d29e6b4', ip: '172.17.0.13', os.name: 'Linux', os.arch: 'amd64', os.version: '3.10.0-229.1.2.el7.x86_64', java.version: '1.7.0_75'

How can I override this?

I even tried, customising node's config.json by adding "host":"192.168.33.10", I get the same error, and log again shows it is still using docker's internal ip?

Any idea why it is still tying to use docker ip instead host ip?

charford:
@sushanghimire I'm assuming you are trying this on the master branch (this has not been merged) ? This pull request would resolve the issue you are describing. Basically, you would add an parameter to the docker run command for the node.

-e REMOTE_HOST="http://1.2.3.4:5555"

The way you are describing it, it sounds like whatever value is being put in remoteHost, is invalid, or blank. If you're doing this on your own branch, make sure you are using the format above(including port) for your -remoteHost parameter.

If there is no remoteHost specified(default behavior), selenium node will randomly choose an interface(from my experience, it's usually the first one listed using ifconfig cmd).

sushanghimire:
@charford, works fine, my issue was related to iptables. This commit should be merged. 

Note this change can be made on config.json as well. But I prefer this approach since it is  unmaintainable to update config.json for each container and it is also against automation principle to change config for each container.

rubytester:
HI, Just started looking at this and I'd like to merge it but I have not personally ever needed to deal with this issue. Any concerns? Missing Tests? Readme addition?

charford:
Still using this setup with a few different servers running nodes, all pointing to a central hub, and have not ran into any issues.

rubytester:
Hello @charford . Some housekeeping. 

Have you [signed the CLA](https://spreadsheets.google.com/spreadsheet/viewform?hl=en_US&formkey=dFFjXzBzM1VwekFlOWFWMjFFRjJMRFE6MQ#gid=0) as described in [CONTRIBUTING.md](https://github.com/SeleniumHQ/docker-selenium/blob/master/CONTRIBUTING.md)?

Please indicate here. Thanks

charford:
@rubytester I signed that last week.

PR #59, opened 2015-05-11
Fix tiny doc typo.

kayabendroth-epages:
The name for the Firefox container to run is very likely meant to be
'fx' and not 'ch' as in the example code for a Chrome container.


willroden:
Thanks.  In addition to fixing the typo, the timing of this also verified that circleci will automatically test pull requests.

Not surprisingly, it appears that changing README.md doesn't break the build.

kayabendroth-epages:
:-)

PR #53, opened 2015-04-13
Standalone Debug dockerfiles for firefox and chrome

rubytester:
- Adding `StandaloneDebug` folder similar to `NodeDebug` but operating on a Standalone Selenium mode.
- Generating `StandaloneDebugChrome` and `StandaloneDebugFirefox` folders from templates
- resulting in docker images `selenium/standalone-chrome-debug` and `selenium/standalone-firefox-debug`

(I have signed CLA in previous contribution to selenium unless I need to sing it again.)

Please let me know if I missed something. 

Thank you.

PR #63, opened 2015-05-23
complete base readme I missed for standalone debug

rubytester:
pull request #4


rubytester:
correction: it was for pull request #53

PR #47, opened 2015-03-04
Added JAVA_OPTS environment variable to control java process.

ipreachable:
#40

PR #64, opened 2015-05-28
Propagate JAVA OPTIONS

mathieu-pousse:
I have updated the entry point to propagate additional parameters as JAVA OPTS for the JVM inside the container.

This is definitely needed to set the http proxy at runtime

I have added a small section in the readme and updated all the entry points to keep it consistent.

### DIFF
--- a/Hub/entry_point.sh
+++ b/Hub/entry_point.sh
@@ -14,7 +14,10 @@ function shutdown {
     echo "shutdown complete"
 }
 
+[ $# -ne 0 ] && echo "Running with JAVA_OPTS = $* " ]
+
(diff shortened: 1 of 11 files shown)


rubytester:
Hi, I would prefer to go with explicit JAVA_OPTS shell var as see in pull request #47 . This way you can add `-e JAVA_OPTS="stuff"` to `docker run` explicitly when building a command to run docker containers.

rubytester:
Hi @mathieu-pousse 
I merged this feature as presented in pull request #47 which allows to be set by explicit JAVA_OPTS env variable. 

Thank you for contributing

Issue #1, opened 2014-11-17
Build with VERSION=local version override with all images


mtscout6:
Initial issue derived from elgalu/docker-selenium#8

The README gives directions to build the images locally by using a local version tag. This unfortunately does not work, because the `FROM` field in each Dockerfile is version specific.

rubytester:
Yep, I see the bug. We'll need generate step to modify the FROM line with the proper VERSION requested when running make.

PR #68, opened 2015-06-09
FROM field in each Dockerfile version specific

rubytester:
Fixes Issue #1

### DIFF
--- a/Hub/Dockerfile.txt
+++ b/Hub/Dockerfile.txt
@@ -1,0 +1,25 @@
+MAINTAINER Selenium <selenium-developers@googlegroups.com>
+
+#========================
+# Selenium Configuration
+#========================
+
+EXPOSE 4444
+
+ENV GRID_NEW_SESSION_WAIT_TIMEOUT -1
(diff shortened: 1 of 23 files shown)


rubytester:
I would appreciate some feedback on this proposed fix. Thanks.

I see the failure in https://circleci.com/gh/SeleniumHQ/docker-selenium/26 is this:

```
http://selenium-release.storage.googleapis.com/2.46/selenium-server-standalone-2.46.0.jar:
2015-06-09 23:09:19 ERROR 503: Service Unavailable.
```

mtscout6:
So the idea is that when it's time for a version bump you will regenerate all the Docerfiles and commit them? Sounds good to me, that will play well with the docker builds!

rubytester:
I can add a `generate_all` target that will change all Dockerfiles apart from generate_x being called before each docker build. This will just change Dockerfiles and you don't have to run docker build on your local machine for all 11 images. 

one more commit then. 


ok, one more commmit with this: 

generate_all

kayabendroth-epages:
+1

PR #85, opened 2015-08-04
Upgrade to Java 8

kayabendroth:


rubytester:
Thanks @kayabendroth 

So we are off LTS and going to 15.04 with Java 8.  

Has anyone had any performance side-effects?

rubytester:
and let's up the chromedriver to 2.18 to support chrome 44

.gitignore
```diff
@@ -4,4 +4,6 @@ node_modules/
 .idea/
 # Sed backup files.
 *-e
+# Vim swap files.
+*.swp
```
mtscout6: You may find this useful :smile: : http://stackoverflow.com/questions/7335420/global-git-ignore

rubytester: I use global `~/.gitignore` for all projects so I don't have to add `.gitignore` to each repo I work on. 

To be pedantic I would gladly remove `.gitignore` file from this repo if most people use `core.excludesfile` directive. 

What say you people ?

mtscout6: I don't think this will hurt anything to have there, I mainly put that there as a note so he doesn't have to add to many repos.

PR #94, opened 2015-09-02
IncreasingMaxInstancesValueForNodes

JeffHall:
Increasing maxInstances and maxSessions values in the config.json files for the Chrome and Firefox nodes.  

Idea here is to take better advantage of the GRID_MAX_SESSION value (5) configured in the Hub, so that as many as 5 browser sessions can run at the same time (either from a single node or distributed across multiple nodes).

### DIFF
--- a/NodeChrome/config.json
+++ b/NodeChrome/config.json
@@ -2,18 +2,18 @@
   "capabilities": [
     {
       "browserName": "*googlechrome",
-      "maxInstances": 1,
+      "maxInstances": 5,
       "seleniumProtocol": "Selenium"
     },
     {
       "browserName": "chrome",
-      "maxInstances": 1,
+      "maxInstances": 5,
       "seleniumProtocol": "WebDriver"
     }
   ],
   "configuration": {
     "proxy": "org.openqa.grid.selenium.proxy.DefaultRemoteProxy",
-    "maxSession": 1,
+    "maxSession": 5,
     "port": 5555,
     "register": true,
     "registerCycle": 5000
--- a/NodeFirefox/config.json
+++ b/NodeFirefox/config.json
@@ -2,18 +2,18 @@
   "capabilities": [
     {
       "browserName": "*firefox",
-      "maxInstances": 1,
+      "maxInstances": 5,
       "seleniumProtocol": "Selenium"
     },
     {
       "browserName": "firefox",
-      "maxInstances": 1,
+      "maxInstances": 5,
       "seleniumProtocol": "WebDriver"
     }
   ],
   "configuration": {
     "proxy": "org.openqa.grid.selenium.proxy.DefaultRemoteProxy",
-    "maxSession": 1,
+    "maxSession": 5,
     "port": 5555,
     "register": true,
     "registerCycle": 5000


smccarthy:
Should it be a static amount or should it be a parameter we can change dynamically? I like using 1 browser instance per container.

rubytester:
I would hold of merging this yet. Here are some thoughts.

This falls under node capacity configuration, a practical need to decorate the node the way you want based on how you run your node with how much CPU power and RAM you have on the machine running a node and I would leave that to the user to provide their own config.json when starting a node.

We could even remove the maxSessions from the hub (I think it will be 5 by default anyway) and we should let the Node dictate maxInstances and maxSessions. I would separate the hub's config from node's config. currently the PR increases maxSessions and maxInstances on arbitrary default value in the hub and if the user increased maxSessions in the hub that hub's setting will not override the node's property value. So number 5 is arbitrary in this situation. 

I think If you would like to decorate the node in a different manner I would suggest using the node as base in your own Dockerfile that consumes this base and then you can provide your own config.json. 

This is actually what I am doing on my project

For example current Firefox is 40. I have created a Dockerfile

```
FROM selenium/node-firefox
USER root
ADD config.json /opt/selenium/config.json
USER seluser
```

and supplied the desired `config.json` based on my VM's used to build the gird.

```json
{
  capabilities: [
    {
      browserName: "firefox",
      version: "40",
      maxInstances: 2,
      seleniumProtocol: "WebDriver"
    }
  ],
  configuration: {
    proxy: "org.openqa.grid.selenium.proxy.DefaultRemoteProxy",
    maxSession: 2,
    register: true,
    registerCycle: 5000
  }
}
```

Basically I decide how to decorate the given docker images in my project's environment based on my capacity.

Of course when Firefox 41 comes I will change your `config.json` and now decorate it as Firefox 41.

I would hold of merging this until getting some feedback. How do people configure the nodes? hub? 

Thoughts?

elgalu:
I personally use `maxInstances: 1` and `maxSession: 1` because some of the tests we run that relies on jQuery :focus fail when 2 browsers are on the same virtual desktop or real desktop, they start competing for the focus.

What works best for me is always running a new selenium container for each test and then dispose the container after that test is done. Bullet proof but maybe overkill, sharing...

JeffHall:
Thanks to all for the great comments!  I've relatively new to Docker, so this PR doesn't really take into account the "Docker mindset" of "1 container for 1 thing" (quoting Mark Collin here).  

I'll withdraw this PR and look into implementing the suggestion offered by @elgalu of starting and stopping node containers per test, which seems like a smart solution.

eyal919:
I agree with Leo about 1 container for one test and I also think this is the way to do automation. but in real life, when you having many tests and you run tests in parallel, you don't always have the resources to create one container to each test.
I think you should give the selenium developer (and the docker user) the option to have more than one instance of browser in the same container. I also think that the developer smart enough to use it as they need.

For example: In my environment, i'm running more than 300 tests in parallel so one instance its just not enough.

SpencerMalone:
You can actually do this at runtime already without having to rebuild your image, @eyal919. Just mount config.json as a volume with the following flag:
-v /local/path/to/config.json:/opt/selenium/config.json

kierr:
I think this should be configurable, rather than arbitrarily upped from 1 to 5. Then that keeps us all happy - both those preferring the single browser per node configuration, and those of us who have many hundreds of concurrent tests to run where a single browser per node setup is illogical, simply based on system resources, based particularly on notoriously resource intensive Java.

PR #100, opened 2015-09-22
Update Dockerfile

sethuster:
Update Chrome Driver Version

### DIFF
--- a/NodeChrome/Dockerfile
+++ b/NodeChrome/Dockerfile
@@ -17,7 +17,7 @@ RUN wget -q -O - https://dl-ssl.google.com/linux/linux_signing_key.pub | apt-key
 #==================
 # Chrome webdriver
 #==================
-ENV CHROME_DRIVER_VERSION 2.18
+ENV CHROME_DRIVER_VERSION 2.19
 RUN wget --no-verbose -O /tmp/chromedriver_linux64.zip http://chromedriver.storage.googleapis.com/$CHROME_DRIVER_VERSION/chromedriver_linux64.zip \
   && rm -rf /opt/selenium/chromedriver \
   && unzip /tmp/chromedriver_linux64.zip -d /opt/selenium \


kayabendroth:
+1

rubytester:
Hello @sethuster 

Some housekeeping: Have you [signed the CLA](https://spreadsheets.google.com/spreadsheet/viewform?hl=en_US&formkey=dFFjXzBzM1VwekFlOWFWMjFFRjJMRFE6MQ#gid=0) as described in [CONTRIBUTING.md](https://github.com/SeleniumHQ/docker-selenium/blob/master/CONTRIBUTING.md)?

Please indicate here. Thanks and we'll merge it.

rubytester:
Can you change the commit message to "added chromedriver 2.19".

I know at some point @elgalu said he just installs the latest chromedriver instead of specifying the version.

I was on the cautious side but I think now that we can change the code to just download the latest chromedriver. 

What do others think please?

kayabendroth:
I would stick to specifying the version we want to use. That way we are in control what a build does and we can expect two builds produce the same thing. A new driver version, at some point, might introduce problems.

elgalu:
I rolled back to [specifying the exact version](https://github.com/elgalu/docker-selenium/blob/95bd73db327a8b6/Dockerfile#L595) as I agree with @kayabendroth. I think the **ideal** solution is to be able to request at docker run time which version we want, e.g.

    docker run selenium -e CHROMEDRIVER=2.19

Or via some json config. And having the last N versions in the image ready to pick one, but that's another ticket...

rubytester:
Thanks for feedback. 

Yeah, the issue I keep having is the grid gives me timeouts on Chrome 44 and Chrome 45. It could be that I am on chromedriver 2.18 or the fact that I run on CentOS 7.1 machines with docker 1.7.1

On docker 1.8.2 on ubuntu VMs I don't have that problem. I will upgrade to Chromedriver 2.19 and we'll see if that helps me with timeouts.

PR #102, opened 2015-09-28
Upgrade Node.js to version 4.1.1.

kayabendroth:


kayabendroth:
The last version before 4.0.0 was 0.12.7. Apparently Node.js merged with io.js and they decided to bump the version like this (https://medium.com/node-js-javascript/4-0-is-the-new-1-0-386597a3436d).

The PR helps keeping the Node dependency up-to-date, but it is not "necessary."

rubytester:
Thank you

PR #104, opened 2015-10-08
allow marathon mesos to set REMOTE_HOST

rubytester:
We spin docker nodes for the hub in Marathon Mesos. 

At the time of starting the docker container Marathon sets MESOS_TASK_ID and we have access to HOST and PORT_5555 env variables from which REMOTE_HOST can be constructed. 

This is an experimental way of spinning nodes. 

Does anyone use Mesos Marathon for their Selenium Grid? How do you deal with this?

### DIFF
--- a/NodeBase/entry_point.sh
+++ b/NodeBase/entry_point.sh
@@ -16,6 +16,10 @@ function shutdown {
   wait $NODE_PID
 }
 
+if [ ! -z "$MESOS_TASK_ID" ]; then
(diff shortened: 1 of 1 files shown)


elgalu:
LGTM

At Zalando we use https://github.com/zalando-stups/senza but the tool right now is more oriented to deploy applications rather than testing infrastructure. To use senza you need a whole setup (private docker registry, oauth2 server and [more](http://stups.readthedocs.org/en/latest/)).

ddavison:
closing as no response given for quite some time

PR #119, opened 2015-11-05
Move ${JAVA_OPTS} so they come before -jar $jarfile in entry_point.sh.

mikewrighton:
The JVM arguments in ${JAVA_OPTS} are being passed in after -jar $jarfile. This, I think, is causing them to be passed to the application as user args, and ignored.

### DIFF
--- a/Hub/entry_point.sh
+++ b/Hub/entry_point.sh
@@ -14,8 +14,7 @@ function shutdown {
     echo "shutdown complete"
 }
 
-java -jar /opt/selenium/selenium-server-standalone.jar \
-  ${JAVA_OPTS} \
+java ${JAVA_OPTS} -jar /opt/selenium/selenium-server-standalone.jar \
   -role hub \
   -hubConfig $CONF &
 NODE_PID=$!
--- a/NodeBase/entry_point.sh
+++ b/NodeBase/entry_point.sh
@@ -25,8 +25,7 @@ fi
 # TODO: Look into http://www.seleniumhq.org/docs/05_selenium_rc.jsp#browser-side-logs
 
 xvfb-run --server-args="$DISPLAY -screen 0 $GEOMETRY -ac +extension RANDR" \
-  java -jar /opt/selenium/selenium-server-standalone.jar \
-    ${JAVA_OPTS} \
+  java ${JAVA_OPTS} -jar /opt/selenium/selenium-server-standalone.jar \
     -role node \
     -hub http://$HUB_PORT_4444_TCP_ADDR:$HUB_PORT_4444_TCP_PORT/grid/register \
     ${REMOTE_HOST_PARAM} \
--- a/NodeChromeDebug/entry_point.sh
+++ b/NodeChromeDebug/entry_point.sh
@@ -27,8 +27,7 @@ fi
 sudo -E -i -u seluser \
   DISPLAY=$DISPLAY \
   xvfb-run --server-args="$DISPLAY -screen 0 $GEOMETRY -ac +extension RANDR" \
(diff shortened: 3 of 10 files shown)


kayabendroth:
@rubytester Can you check the deploy key settings in CircleCI? There was a "Permission denied" error.

kayabendroth:
+1

mathieu-pousse:
+1

rubytester:
About Circle CI: I don't think I have access to that setup... and they had some issues  with builds http://status.circleci.com/

About moving the JAVA_OPTS. This came from PR #47. My understanding was that we could use it to pass some flags to the selenium, e.g "-debug" so it would start selenium in debug mode.

if we move the JAVA_OPTS before the -jar do we lose that option?

mikewrighton:
PR #47 seems to be about JVM options - there is a comment about adding ```-Xmx1024m```, for example. (fwiw I think being able to control the heap is essential for this project.)

Maybe there should be another variable for adding extra Selenium options?

Regarding CircleCI, I've never used it before but I was playing around with my account settings and from what I can tell it's now building successfully. Although I don't see the results in here...

apakhomov:
I think it'll be useful to split jvm and selenium opts into 2 vars

rubytester:
ok, we can split the two. I am curious what usecases people have and what would work for you

apakhomov:
We always increase Xmx, as grid begins to complain with default settings at 500+ working browsers. Also I prefer to pass settings (timeouts, etc) to grid via cli parameters - in this way you can see startup settings in process list

kayabendroth:
+1 for having two OPTS variables. Makes sense.

rubytester:
Back to splitting the jvm and selenium opts into 2 vars. `JAVA_OPTS` moves before jarfile and we should introduce a new env var: Any name suggestions? Usage examples? 

`JAR_OPTS` ? 

`SE_OPTS` ? 

@mikewrighton would you be willing to add that to this PR ?

mikewrighton:
@rubytester sure, will update the PR shortly.

rubytester:
I apologize to all. When PR #47 arrived I assumed that adding the opts at the end is fine and position did not matter as long as the args were on the line, in that I was ignorant. Time to fix it then. Thanks for the eduction on this.

I suggest the usage of `SE_OPTS` then as a secondary env so you can add `-debug` if you wanted to or modify default options already assigned for the hub and the node).

I will add some code to this PR for review.

rubytester:
add SE_OPTS on top of this PR https://github.com/SeleniumHQ/docker-selenium/commit/0ecdabde7d3d1b01501b59b09c2c3ec69023f4f7

Would this be ok?

mikewrighton:
@rubytester Looks good to me.

rubytester:
something with Circle CI went to borked... digging now.

rubytester:
hi @mikewrighton 
Did you/could you sign the CLA please according to this doc: https://github.com/SeleniumHQ/docker-selenium/blob/master/CONTRIBUTING.md to accept your patch. 
Please indicate here.
Thanks.

rubytester:
Additional work done on PR #124

mikewrighton:
@rubytester I just signed it. Thanks for your help.

kayabendroth:
@rubytester You can create a deploy key in CircleCI via Project Settings >> Permissions >> Checkout SSH keys. I think after you click on Create and add <project> deploy key a read-only deploy key will be generated and stored. This key can be used for reading, but not for writing to the repository, but this should be enough here.

If you check the settings for this project, you should see a deploy key entry named "SeleniumHQ/docker-selenium deploy key."

PR #121, opened 2015-11-13
JAVA_OPTS was in wrong location, doesn't work

jskjons:


jskjons:
oops, it looks like this is a duplicate of #119

PR #131, opened 2015-12-03
manullay transmit env variables to selenium user in *debug*

mathieu-pousse:
the debug images are using the `sudo -E -i -u seluser` to start the selenium stuffs and browser with the appropriated seluser.

Unfortunatly, as described in the sudoers documentation: 

> As a special case, if sudo's -i option (initial login) is specified, sudoers will initialize the environment regardless of the value of env_reset. 

That means the env variables from the docker client won't be transmitted to the browser with `-i`

As the `-i` is mandatory to initiate the browser stuffs and `-E` is useless, I have added a trick to extract env variables from the root users and pass them to the seluser.

This is definitly needed to make the debug images working

### DIFF
--- a/NodeChromeDebug/entry_point.sh
+++ b/NodeChromeDebug/entry_point.sh
@@ -28,7 +28,10 @@ fi
 
 # TODO: Look into http://www.seleniumhq.org/docs/05_selenium_rc.jsp#browser-side-logs
 
-sudo -E -i -u seluser \
+env | cut -f 1 -d "=" | sort > asroot
+  sudo -E -u seluser -i env | cut -f 1 -d "=" | sort > asseluser
+  sudo -E -i -u seluser \
+  $(for E in $(grep -vxFf asseluser asroot); do echo $E=$(eval echo \$$E); done) \
   DISPLAY=$DISPLAY \
   xvfb-run --server-args="$DISPLAY -screen 0 $GEOMETRY -ac +extension RANDR" \
   java ${JAVA_OPTS} -jar /opt/selenium/selenium-server-standalone.jar \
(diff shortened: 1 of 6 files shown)


rubytester:
I have never run into this before so I can't really provide any good feedback on this. Is the reason for the change because we drop some values? Anybody else please?

mathieu-pousse:
Yes, I must provide the http_proxy and https_proxy in my environment and those variables are lost in "debug mode"

The way to reproduce it in a similar env is the following: 

```bash
# start a container with environment variable ANYTHING
$ docker run -it -e ANYTHING=123 ubuntu:14.04 bash

# it is available
root@f29746d3ae3d:/# env | grep ANYTHING
ANYTHING=123

# create a user seluser as in docker-selenium images
root@f29746d3ae3d:/# adduser seluser
Adding user 'seluser' ...

# sudo w/o any option: ANYTHING is not propagated
root@f29746d3ae3d:/# sudo -u seluser env | grep ANYTHING

# sudo w -E option (to preserve the current environment variables in the child bash): ANYTHING is propagated
root@f29746d3ae3d:/# sudo -E -u seluser env | grep ANYTHING
ANYTHING=123

# sudo w -E -i option (to preserve the current environment variables in the child bash but also run .bashrc (needed to start a browser)): ANYTHING is not propagated
root@f29746d3ae3d:/# sudo -E -i -u seluser env | grep ANYTHING
```

The `-i` option drop all the env variables from the parent, regardless of other options as `-E`

The only way I found to make it work is here in my commit

HIH

rubytester:
Thanks for the explanation. 2 things.

1. Can you rebase this on master and resolve conflicts (not ready to merge yet)
2. Have you signed the CLA please according to this https://github.com/SeleniumHQ/docker-selenium/blob/master/CONTRIBUTING.md ?  Please indicate here. Thanks.

mathieu-pousse:
here we are, yes i've signed the CLA

thx

mathieu-pousse:
Well done! Could you tell me if the CI does push the new docker images on the docker hub every time there is a commit on the master?

This is to know when I could rely on official images instead of using the one from our fork.

Thanks in advance for the response and for what you do with OSS !

M.

PR #136, opened 2015-12-16
Add HUB_HOST environment variable for Docker networking clarity

### LINKED ISSUE #133: Use Docker networking instead of links
> Docker links are a one-way, single-host communication system. They should now be considered deprecated, and you should update your app to use networking instead. 

See https://docs.docker.com/compose/networking/

There is [a condition in NodeBase](https://github.com/SeleniumHQ/docker-selenium/blob/master/NodeBase/entry_point.sh#L9) that causes node containers to exit when networking is enabled, because it causes links to be ignored.

SpencerMalone:
Helps resolve #133 by increasing clarity on how to use the new (Docker v1.7) networking feature.

### DIFF
--- a/NodeBase/entry_point.sh
+++ b/NodeBase/entry_point.sh
@@ -6,6 +6,11 @@ if [ ! -e /opt/selenium/config.json ]; then
   exit 1
 fi
 
+if [ ! -z "$HUB_HOST" ]; then
+  echo "Connecting to the selenium hub using the host string ${HUB_HOST}"
+  HUB_PORT_4444_TCP_ADDR=$HUB_HOST
+fi
+
 if [ -z "$HUB_PORT_4444_TCP_ADDR" ]; then
   echo Not linked with a running Hub container 1>&2
   exit 1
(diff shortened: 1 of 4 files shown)


rubytester:
I would hold off on this. Can we gather some more feedback per #133 discussion please? It may be that in light of new `docker network` we should expose env variable that were previously hidden by the `--link` and yes, a bit ugly.  We can go with `HUB_HOST` and `HUB_HOST_PORT` instead of those cryptic ones auto assigned by `--link`.

rubytester:
I do see a need to add to README the bits about configuring the grid on multi host machines. Basically Case 1 is covered, now Case 2 and Case 3 scenario can be either added to README or we can start adding pages in wiki section.

vivanov1410:
ping

diemol:
I'll close this one since most of its content is covered by #640. I am not sure why this one didn't came through, but anyway thanks to @SpencerMalone and @rubytester for providing most of the content used in #640.

PR #170, opened 2016-02-12
Add StandaloneChrome ImageLayers badge

Vanuan:
Let's add some ImageLayers badges.


kayabendroth:
@Vanuan Can you add this for all of our images and add it to this Pull Request?

PR #171, opened 2016-02-12
Add Base ImageLayers badge

Vanuan:
Let's add some ImageLayers badges.


kayabendroth:
Change should be added to https://github.com/SeleniumHQ/docker-selenium/pull/170.

PR #174, opened 2016-02-24
Document how to access host URL's from within the container

ShaunDychko:
When developing with a web server running on the host, the browser within the container needs to access the URL that's available only within the host. Adding `--net=host` to the `docker run` command accomplishes this.

See https://docs.docker.com/v1.8/articles/networking/#container-networking for more details, including a warning that "You should use this option with caution." Should a warning like that be included here? When used in a gitlab test runner, for example, the option shouldn't be a problem, but let me know if this PR should be modified to mention that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,9 @@ You can pass `SE_OPTS` variable with additional commandline parameters for start
 $ docker run -d -p 4444:4444 -e SE_OPTS=-debug --name selenium-hub selenium/hub:2.52.0
 ```
 
+### Connect to host network from container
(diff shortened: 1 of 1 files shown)


diemol:
Hi @ShaunDychko,

Thanks for the PR.  And sorry for the super late review.

This documentation is only valid for Linux, would you care to document it also for Mac and Windows? I think it will be easier if you add this as a Wiki entry https://github.com/SeleniumHQ/docker-selenium/wiki

I'll close the PR and hopefully you still have the enthusiasm to create a Wiki entry.

PR #177, opened 2016-03-15
Update Dockerfile to use wily apt repo

dcosson:
Not sure if this was an oversight or an intentional workaround for something, but 15.10 is wily, not vivid. When trying to extend the standalone base docker image I got some failures due to unexpected versions.

Vivid is also past EOL, so there might be security issues from using the old packages (e.g. the glibc getaddrinfo bug was more recent than vivid EOL).

### DIFF
--- a/Base/Dockerfile
+++ b/Base/Dockerfile
@@ -4,8 +4,8 @@ MAINTAINER Selenium <selenium-developers@googlegroups.com>
 #================================================
 # Customize sources for apt-get
 #================================================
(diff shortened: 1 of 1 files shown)


ti2m:
I ran into issues when installing fonts, took me a while to figure out why installs failed as I didn't catch the release name issue right away. Updating the repos fixed the issues for me.

kayabendroth:
@dcosson This was not intentional. Fixed with https://github.com/SeleniumHQ/docker-selenium/pull/199.

PR #181, opened 2016-03-17
chromedriver https download and version update

dcosson:
Use https for downloading chromedriver since we don't check signature or sha, and update to the latest version.


kayabendroth:
Superseded by (and merged with) https://github.com/SeleniumHQ/docker-selenium/pull/194.

PR #187, opened 2016-03-22
Fix setting of server-num in xvfb-run, fixes #184

### LINKED ISSUE #184: xvfb-run is being used incorrectly to set DISPLAY
The current use of `xvfb-run` with respect to `DISPLAY` is:
```
xvfb-run --server-args="$DISPLAY -screen 0 $GEOMETRY -ac +extension RANDR" \
```

The applicable documentation for `xvfb-run` is:
```
-n NUM    --server-num=NUM          server number to use (default: 99)
-s ARGS   --server-args=ARGS        arguments (other than server number and
                                    "-nolisten tcp") to pass to the Xvfb server
                                    (default: "-screen 0 640x480x8")
```

Therefore, adding the `DISPLAY` into `--server-args` is incorrect. If I create a container and override the `DISPLAY` environment variable, `xvfb` won't start.

In my use case I am putting the hub and nodes in the same network, so using the same `DISPLAY` isn't working. The current containers work because `DISPLAY=:99.0` which is the default for `xvfb-run`.

double16:
See issue #184 for a full description of the problem.

### DIFF
--- a/NodeBase/entry_point.sh
+++ b/NodeBase/entry_point.sh
@@ -28,7 +28,8 @@ fi
 
 # TODO: Look into http://www.seleniumhq.org/docs/05_selenium_rc.jsp#browser-side-logs
 
-xvfb-run --server-args="$DISPLAY -screen 0 $GEOMETRY -ac +extension RANDR" \
+SERVERNUM=$(echo $DISPLAY | sed 's/:\([0-9][0-9]*\).[0-9][0-9]*/\1/')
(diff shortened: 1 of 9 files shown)


kayabendroth:
Superseded by https://github.com/SeleniumHQ/docker-selenium/pull/206. @double16 Please check my enhancements on top of your commit. Will merge the other PR, if you agree.

double16:
Your enhancements look good, I like the extraction into functions.sh.

PR #199, opened 2016-04-08
Updated the apt sources to use "wily" (15.10) instead of "vivid" (15.04).

charford:
I ran into issues trying to install some stuff using apt-get, and it turned out to be because the sources file was still using sources for 15.04. I have verified that using the correct sources, the make command still builds all docker images successfully.


charford:
Fixes #200

kayabendroth:
Thanks.

PR #204, opened 2016-04-14
RUN commands already run as root, no need to sudo | tee here.

karel1980:
Minor simplification


kayabendroth:
@karel1980 You're right. And while already touching those lines, I used the content of $TZ in the echo call.

PR #217, opened 2016-05-04
Add mp4 playback ability in firefox for html5

Z3r0Sum:
## Problem
- Address inability to play mp4 video within html5 for firefox.
## Solution
- Add installation of the `gstreamer1.0-libav` and its subsequent dependencies for `NodeFirefox`.
## Testing
- [x] Build and launch a firefox container that connects to a Selenium Grid Hub container from this project.
- [x] Launch Selenium tests that required mp4 video playback capabilities. 
- [x] Tests Pass.

I could potentially roll this fix into the docker `RUN` command for firefox, but it was broken out for clarity.  If you would like me to include it within that call, please let me know.

### DIFF
--- a/NodeFirefox/Dockerfile
+++ b/NodeFirefox/Dockerfile
@@ -18,6 +18,12 @@ RUN apt-get update -qqy \
   && mv /opt/firefox /opt/firefox-$FIREFOX_VERSION \
   && ln -fs /opt/firefox-$FIREFOX_VERSION/firefox /usr/bin/firefox
 
+#=============================
+# Install mp4 video libraries
+#=============================
+RUN apt-get update -qqy \
+  && apt-get -qqy --no-install-recommends install gstreamer1.0-libav
+
 #========================
 # Selenium Configuration
 #========================
--- a/NodeFirefox/Dockerfile
+++ b/NodeFirefox/Dockerfile
@@ -8,7 +8,9 @@ USER root
 #=========
 ENV FIREFOX_VERSION 45.0.2
 RUN apt-get update -qqy \
-  && apt-get -qqy --no-install-recommends install firefox \
+  && apt-get -qqy --no-install-recommends install \
+       gstreamer1.0-libav \
+       firefox \
   && rm -rf /var/lib/apt/lists/* \
   && wget --no-verbose -O /tmp/firefox.tar.bz2 https://download-installer.cdn.mozilla.net/pub/firefox/releases/$FIREFOX_VERSION/linux-x86_64/en-US/firefox-$FIREFOX_VERSION.tar.bz2 \
   && apt-get -y purge firefox \


Z3r0Sum:
@kayabendroth Any idea when this will be looked at?  Just trying to get an estimate.

elgalu:
What about adding some automated tests? That prove:

> Launch Selenium tests that required mp4 video playback

Note I [included](https://github.com/elgalu/docker-selenium/commit/576f0856ab1b664ce430cb0428ded1800ef4eeca#diff-4ac32a78649ca5bdd8e0ba38b7006a1eR13) your change in the other docker-selenium project, see [here](https://github.com/elgalu/docker-selenium/blob/2.53.0o/Dockerfile#L210) and nothing broke ;)

NodeFirefox/Dockerfile
```diff
@@ -18,6 +18,12 @@ RUN apt-get update -qqy \
   && mv /opt/firefox /opt/firefox-$FIREFOX_VERSION \
   && ln -fs /opt/firefox-$FIREFOX_VERSION/firefox /usr/bin/firefox
 
+#=============================
+# Install mp4 video libraries
+#=============================
+RUN apt-get update -qqy \
+  && apt-get -qqy --no-install-recommends install gstreamer1.0-libav
```
kayabendroth: Please add `gstreamer1.0-libav` to line 11 and break it up like this:

```
&& apt-get -qqy --no-install-recommends install \
        gstreamer1.0-libav \
        firefox \
&& ...
```

This saves another layer *and* a call to `apt-get update`.

Z3r0Sum:
Yeah, I'll make those changes tomorrow and try to write some corresponding tests.

Z3r0Sum:
@elgalu What's the best way to do this?  Adjust the `smoke-test.js` and add something that hits html5 video via `browser.get`?  I've never written any javascript before, so any pointers would be helpful.  Thanks!

elgalu:
Yes that one is a different project, in this project the proper place would be the firefox images only. Idk how to add the html5 video tests without doing some research which I'm gladly leaving that to you ;)

Z3r0Sum:
@elgalu Should it be done within that JS `smoke-test` though?

elgalu:
Idk really, better ask @kayabendroth

jmbowman:
It looks like this timed out before it got cleaned up and merged; would a newer PR for this with tests be welcome, or is mp4 playback considered out of scope for these images?  I'm switching over an existing test suite from Vagrant in which many tests need the browser to support mp4 playback, and I suspect that it's not that uncommon of a requirement.  On the other hand, adding this package and all of its dependencies would increase the image's disk space by about 122 MB (it's 32.4 MB worth of Ubuntu package downloads).

In the short term, I'll probably create my own image extending this one to handle it (I've considered using the elgalu images, but they're even larger by far). Just wondering if it's also worth submitting a PR for it.

Issue #216, opened 2016-05-04
connecting through VNC documentation issue


vikramvi:
For first time user who is reading the documentation related to VNC connection there are few confusing points , please review and enhance
1. vnc://<boot2docker-ip>:49160 instead of hard coding  49160 , it should be mentioned as <port4VNC> value
2. <port4VNC>:5900 this is not clear as one point it says "...5900 is fine if it is free, but.."
   but it also mentions
   $ docker run -d -P -p 5900:5900 --link selenium-hub:hub selenium/node-chrome-debug:2.53.0
   $ docker run -d -P -p 5901:5900 --link selenium-hub:hub selenium/node-firefox-debug:2.53.0

this part needs more info to understand for new comers.

thanks & regards,
Vikram

zypherman:
I am not sure that your second point really needs clarification. If you understand what the port flag means then it should be clear what the documentation is saying. All it's saying is that the vnc service is running on 5900 on the node but externally you can map it to whichever port you please.

thataustin:
On this topic, is the `--link` part necessary to get the VNC server running correctly?

vikramvi:
@thataustin Thanks for asking this query.

This proves that documentation is not clear enough for all the audiences. IMO any technical documentation should be written in as simple english as possible without "assuming" all are experts in networking and/or selenium usage. cc: @zypherman

ddavison:
if there is an improvement you can make to the documentation - please! submit a pr! :)

zypherman:
@thataustin The --link flag is to enable the hub and node to talk to each other correctly when they are both running on the same machine. It links the containers together to establish that communication. For the VNC part its not really directly related since you can have a standalone node running and connect to the VNC through that.

diemol:
Closing this issue as since the doubts were solved by the community. If something is missing from the documentation, please send us a PR and become a contributor :)

PR #221, opened 2016-05-23
Clarifying documenting around port mapping for the VNC service

zypherman:
https://github.com/SeleniumHQ/docker-selenium/issues/216
As per this issue I have adding some clarification around the port mapping for the VNC service on the debug nodes. The current documentation is a tad confusing at first but these additions should help make it clearer whats happening.


ddavison:
thanks @zypherman , please also make sure you've signed the [CLA](http://goo.gl/qC50R), and for future pull requests, make sure that your `git config --global --list` is up to date. looks like your `user.email` could be incorrect?

Issue #169, opened 2016-02-12
iron.io microcontainers


Vanuan:
Currently, selenium images are huge (700 mb)
Probably, it is possible to shrink that down:
http://www.iron.io/microcontainers-tiny-portable-containers/

Vanuan:
Selenium base images is built on top of ubuntu:15.10
Let's try to replace it with alpine:3.3 (replacing `apt-get install` by `apk add`) and see what happens.

Vanuan:
Oh, this looks to be a duplicate of #161

Vanuan:
[Base image](http://imagelayers.iron.io/?images=selenium%2Fbase:latest) consists of ubuntu (134 mb) and jre (320 mb) and selenium itself (31 mb).

Vanuan:
[Alpine + minimal jre](http://imagelayers.iron.io/?images=iron%2Fjava:latest) is 111 mb, saving us 375 mb!

nisaacson:
+1, dockerhub is moving towards alpine linux as well for the base library images https://www.brianchristner.io/docker-is-moving-to-alpine-linux/

jgielstra:
One issue is alpine uses `musl` for it's c library and it isn't 100% compatible with stuff compiled using `glibc`.

They do have ports of xvfb, chromium, openjdk8, firefox, and chromium-chromedriver  _webdriver_ in their main repo.
  
It doesn't seem stable and I get <defunct> processes but it looks like people are working on it  
     http://git.alpinelinux.org/cgit/aports/log/community/chromium

But something basic: openjdk + chromium + xvfb + chromium webdriver is ~350 megs..

Vanuan:
Yeah, as I wrote, we only save 375 mb.

ddavison:
closing in favor of #224

Issue #161, opened 2016-02-04
Alpine Linux as base image


lcacciagioni:
Guys how about using alpine linux as the base image? This can be a great enhancement in the reduction of the final size of the image.

andrerom:
+1 This is only image I personally use that still pulls in ubuntu, so switching to jessie, or better yet jumping directly to alpine linux like others offical images are about to would be great to reduce build/test/download time for everyone.

progre55:
+1! Can't wait for this!

danielwhatmuff:
+1

elgalu:
With github is now possible to +1 via "reactions" please avoid polluting issues with +1 comments, see:

[![IMAGE ALT TEXT HERE](https://cloud.githubusercontent.com/assets/25792/13683303/d55b8b7e-e6ba-11e5-9b31-59a8db92e456.gif)](https://github.com/blog/2119-add-reactions-to-pull-requests-issues-and-comments)

ylavoie:
I was looking for that too. I'm almost done, Hub & Firefox are working on Alpine, Chrome & PhantomJS will soon. I will PR my proposal.

andrerom:
@ylavoie Found a way around issues in #224, or are they solved in alpine now?

diemol:
Hi all, 

I just checked the packages for Chromium and Firefox, e.g.:
Chromium:
- https://pkgs.alpinelinux.org/package/v3.7/community/x86_64/chromium
- Latest release is 61.0.3163.100-r0, almost more than one month ago and the current Chrome version is >=63
Firefox:
- https://pkgs.alpinelinux.org/package/edge/testing/x86_64/firefox
- Is a bit more up to date, latest release 57.0.1 (current release is 57.0.2)

Even if the other dependencies are solved (like x11vnc), looks like the browser releases are way behind the releases from the browser vendor channels. We could be potentially blocked to use new browser versions since the new version package is not there yet for Alpine.

I would love to have smaller images, of course. But for me it seems that the all the effort plus having browsers not up to date is not really compensating the benefit of having smaller images.

What do you all think?

ddavison:
not to mention, @diemol, Chromium != Google Chrome.   Users of Selenium might find that alarming if they want to test "Google Chrome" and get "Chromium"

diemol:
Right, forgot to mention that as well @ddavison, you are totally right.

I'll close this issue based on the feedback gotten here and on #224

PR #224, opened 2016-05-25
:construction: Add alpine variant.

shouze:
It's a PR for a early review. Ref to #161 & #169.

### DIFF
--- a/NodeBase/alpine/Dockerfile
+++ b/NodeBase/alpine/Dockerfile
@@ -10,15 +10,12 @@ USER seluser
 
 RUN sudo apk add --no-cache \
         xvfb \
-        tzdata \
         alpine-conf \
-    && sudo setup-timezone -z "US/Pacific" \
     && sudo chmod +x /opt/bin/entry_point.sh
 
 ENV SCREEN_WIDTH=1360 \
     SCREEN_HEIGHT=1020 \
     SCREEN_DEPTH=24 \
-    DISPLAY=:99.0 \
-    TZ="US/Pacific"
+    DISPLAY=:99.0
 
 CMD ["/opt/bin/entry_point.sh"]
--- a/NodeBase/alpine/Dockerfile.txt
+++ b/NodeBase/alpine/Dockerfile.txt
@@ -9,15 +9,12 @@ USER seluser
 
 RUN sudo apk add --no-cache \
         xvfb \
-        tzdata \
         alpine-conf \
-    && setup-timezone -z "US/Pacific" \
     && sudo chmod +x /opt/bin/entry_point.sh
 
 ENV SCREEN_WIDTH=1360 \
     SCREEN_HEIGHT=1020 \
     SCREEN_DEPTH=24 \
-    DISPLAY=:99.0 \
-    TZ="US/Pacific"
+    DISPLAY=:99.0
 
 CMD ["/opt/bin/entry_point.sh"]


shouze:
Here is the early result in term of size of the first container builds:
``` 
selenium/node-base    2.53.0-alpine   151.3 MB
selenium/node-base    2.53.0          371.6 MB
selenium/base         2.53.0-alpine   133.9 MB
selenium/base         2.53.0          293.3 MB
```

stephpy:
Awesome ! 👍

kayabendroth:
@shouze Regarding the TZ setting: I don't know. But I would like to have this TZ change, if we want to do it, in another PR.

kayabendroth:
@shouze I like the idea to have an Alpine variant, not interfering with the Ubuntu based images. On the other hand: all those images. What do the others think?

Alpine used to have issue with DNS resolution (https://github.com/smebberson/docker-alpine/issues/24, https://support.tutum.co/support/discussions/topics/5000048360).

shouze:
@kayabendroth yes I think Alpine have to exist as a side variant, at least during a mid term transition phase (or not, depending  if you wish to maintain Alpine & Ubuntu variants on the long term).

About the DNS issue I think it's ok now as libmusl was missing some feature at some time.
I can continue to rollout this PR if you whish.

As a next step, maybe I can make try to make all chrome containers for example and add some CI tests. Or maybe you have suggestions on the way to make the next move on this PR?

mtscout6:
I believe the timezone setting was an artifact from @elgalu's original work at the beginning of this project. I'd imaging removing it should be just fine and just use UTC.

shouze:
@mtscout6 ok done, @elgalu can you plz confirm when you have some time?

elgalu:
@shouze I suppose is safe to remove `TZ` here from the Docker files, I'll stick with `TZ` [default to Berlin](https://github.com/elgalu/docker-selenium/blob/2.53.0p/Dockerfile#L92) in the other project.

roydekleijn:
any update on alpine ?

shouze:
Didn't work on IT since but feel free to contribute on this PR if you're interested And have some free time ;)

ddavison:
I think for the 3.0.0 release, I will be, instead of variants, make the base alpine to start with.  that way we don't have to manage different base images

ddavison:
so, i've been working on converting us to alpine; here are my findings...

1. x11vnc is not available to install from [main](http://dl-cdn.alpinelinux.org/alpine/edge/main/x86_64/) or [community](http://dl-cdn.alpinelinux.org/alpine/edge/community/x86_64/) with `apk add`. 
2. All alpine executables are compiled by the Alpine community and provided.  Both Chrome and ChromeDriver are proprietary, and follow a different versioning system.  This means that we can't specify `ARG CHROME_DRIVER_VERSION=2.24`.  it would have follow [their schemes](https://pkgs.alpinelinux.org/package/edge/community/x86_64/chromium-chromedriver) which could be very confusing.
3. Alpine provides "chromium."  This is the base chromium browser - not Google Chrome, therefore I don't believe we will be able to use Google Chrome in the alpine images.

Now, #1 and #2 are solvable.   1)  we can manually copy a x11vnc.apk package into the containers.  2) we could use their source to build, but could take a while to even build the docker images, and not sure about that.

SUNx2YCH:
@ddavison, regarding `#3`, I played a bit with `chromium` for my pet project ([SUNx2YCH/docker-alpine-selenium](https://github.com/SUNx2YCH/docker-alpine-selenium)), but couldn't get it to work. There were some errors related to GPU:
```
[544:544:1014/092803:ERROR:gl_surface_glx.cc(381)] glxQueryVersion failed
[544:544:1014/092803:ERROR:gl_initializer_x11.cc(20)] GLSurfaceGLX::InitializeOneOff failed.
[544:544:1014/092803:ERROR:gpu_child_thread.cc(390)] Exiting GPU process due to errors during initialization
[338:360:1014/092803:ERROR:browser_gpu_channel_host_factory.cc(119)] Failed to launch GPU process.
```
Fast googling didn't help. They say that similar issue was already fixed in v50 (mine is v53.0):
https://bugs.chromium.org/p/chromedriver/issues/detail?id=1298
So it's not an easy option even with pre-compiled packages. JFYI.

ddavison:
I appreciate you tuning in @SUNx2YCH and giving your findings.  I don't know what to say.. I think we may have to keep ubuntu as base and just take the "+100MB more" of disk space for using ubuntu versus Alpine.

Unless other people can tune in?

I'll leave this issue open for discussion.

andrerom:
Well if alpine is out of reach for now, make sure to report issues upstream, track them here. And maybe switch to Debian for defaults like official images so most users don't have to download an additional distro image.

ckortekaas:
@ddavison it looks like x11vnc has recently (https://git.alpinelinux.org/cgit/aports/log/unmaintained/x11vnc?showmsg=1) moved from unmaintained to get a maintainer and is available in testing here: http://dl-cdn.alpinelinux.org/alpine/edge/testing/x86_64/x11vnc-0.9.14-r2.apk

And I suspect a majority of people would be happy testing in an alpine chromium instead of google chrome with the reduced file size (and resulting continuous integration drop in build times).

It looks like others are building their own alpine-selenium-chromium packages to work around this in the mean time, but there are many benefits to sticking with your well supported SeleniumHQ builds.

Does the x11vnc change make alpine selenium worth pursuing or is it still out of reach?

diemol:
I just checked the packages for Chromium and Firefox, e.g.:
Chromium:
- https://pkgs.alpinelinux.org/package/v3.7/community/x86_64/chromium
- Latest release is 61.0.3163.100-r0, almost more than one month ago and the current Chrome version is >=63
Firefox:
- https://pkgs.alpinelinux.org/package/edge/testing/x86_64/firefox
- Is a bit more up to date, latest release 57.0.1 (current release is 57.0.2)

Even if the other dependencies are solved (like x11vnc), looks like the browser releases are way behind the releases from the browser vendor channels. We could be potentially blocked to use new browser versions since the new version package is not there yet for Alpine.

I would love to have smaller images, of course. But for me it seems that the all the effort plus having browsers not up to date is not really compensating the benefit of having smaller images.

What do you all think?

shouze:
Yup, pretty convinced that this issue should be closed, not that easy to produce an alpine variant.

andrerom:
Maybe this project should then rather then align with newer official images on usage of debian:stretch-slim instead? _(smaller, and avoids loading several base images for most users)_

diemol:
I'll close this PR based on the feedback gotten here and on #161 

@andrerom it is not a bad idea to base it in a different image as long as the current features still work and we also save space. It would be cool if someone gives us a hand with that. For now I think we should work on improve the images to support other languages and find out if we really need the debug and normal images (perhaps we can merge them if the performance and CPU/memory usage are very similar).

tnguyen14:
@diemol our project just ran into an issue where screenshots generated on debug and normal images are different (when feeding to a Visual Regression test). We're unsure as to why, but decided to just standardize on using debug images, just to have it consistent. The ability to optionally watch what's going on is too important for us to standardize on normal images.

I think Chrome itself already uses significant memory anyway, so I think the idea of merging the two would be interesting.

PR #225, opened 2016-05-26
Upgrade FF to 46.0.1

Skool:
Some of our jobs didnt work with 45.0.2 but does with 46.0.1 who is the last current stable version of FF.


Skool:
I've run test.sh in local, and I got no errors. How to debug the problem ?

Issue #91, opened 2015-08-21
node-chrome-debug:2.47.1 xvfb-run: error: Xvfb failed to start


MarceloEmmerich:
I have the following setup in a docker-compose.yml:

```
hub:
  image: selenium/hub
  ports:
    - "4444:4444"
chrome:
  image: selenium/node-chrome-debug:2.47.1
  ports:
    - 5900
  links:
    - hub
    - nginx
```

when I launch my test, I often receive the following error from the logs of the chrome container:

``xvfb-run: error: Xvfb failed to start``

and my test runner (nightwatchjs) says:

``
Error retrieving a new session from the selenium server
{ status: 13,
  value: 
   { class: 'org.openqa.grid.common.exception.GridException',
     message: 'Error forwarding the new session cannot find : Capabilities [{platform=ANY, acceptSslCerts=true, javascriptEnabled=true, browserName=chrome, name=Evaluator Login, chromeOptions={args=[--no-sandbox, --window-size=1400,1100, --lang=de, --disable-web-security]}}]' } }
``

When I connect with VNC to the container previous to running the tests, it works 100% of the time. However, I want this to be part of the CI process, so it also has to run unattended. Any ideas? Thanks, Marcelo

rubytester:
I see you run a debug node attached to the hub. Maybe there are timing issues in spinning up a hub and then adding a node... 

A better way may be to start a `standalone-chrome-debug` image if you are running a remote driver on demand. That way you bypass the entire grid. I assume you only need one browser session. 

The hub plus nodes is a good setup when you need to spin 20 nodes for example because you need to run parallel tests.

kingkevin:
I started running into this issue as well using the node-chrome docker image. The xvfb error is triggered intermittently.

pwaller:
I've just hit this, I happened to notice it started when I upgraded to docker 1.8.1 from docker 1.7.1.

pwaller:
Okay, after a bit of fun trying to reproduce and bisect, I've discovered why it's happening. It's actually my docker-compose upgrade that did it, not docker's upgrade. I can reproduce it on both 1.7 and 1.8.

The difference is with docker compose 1.4, if you don't use `--force-recreate`, then sometimes it will get stuck saying `Waiting xvfb...`.

Now that I know this, I'm able to reproduce it without docker-compose.

In one window, run this:

```
$ docker run -ti --rm --name hub selenium/hub
```

Then in another:

```
[...]

PR #230, opened 2016-06-01
Seeing intermittent failures on restart of nodes due to a failure to …

scottturley:
…start xvfb.

The issue is documented here:

https://github.com/SeleniumHQ/docker-selenium/issues/91

Looks to be a problem with the lock files preventing a restart intermittently. Adding a line here to remove the lock files on startup.

### DIFF
--- a/NodeBase/entry_point.sh
+++ b/NodeBase/entry_point.sh
@@ -32,6 +32,9 @@ fi
 # TODO: Look into http://www.seleniumhq.org/docs/05_selenium_rc.jsp#browser-side-logs
 
 SERVERNUM=$(get_server_num)
+
+rm -f /tmp/.X*lock
+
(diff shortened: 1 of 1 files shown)


anx-ckreuzberger:
I've been using this as well for a bit, it's really great because now I don't have tore-create the docker instances.

kostajh:
FWIW, we are seeing the same issue with the standalone chrome debug image, so perhaps https://github.com/SeleniumHQ/docker-selenium/blob/master/StandaloneChromeDebug/entry_point.sh needs to get updated as well?

ddavison:
committed as e116033037e7f9b91bea81a2ffbe723572d079dd.  thanks!  also, have you signed the [CLA](https://spreadsheets.google.com/spreadsheet/viewform?hl=en_US&formkey=dFFjXzBzM1VwekFlOWFWMjFFRjJMRFE6MQ#gid=0)?

PR #243, opened 2016-07-09
Update README.md

juliangamble:
Handle external files traversing a Corporate Proxy.

Increasingly common with Corporate VPC setups going via the intranet proxy to acces files hosted externally.


ddavison:
apologies for nobody getting to you @juliangamble, unfortunately this case is extremely difficult to reproduce, and i'm not sure anybody else has the same issue you have so there's really no way to validate what you are saying.

would you mind if i just add this pull request as a new wiki page instead, with a disclaimer?

juliangamble:
That's fine. Just as long as the solution is available on a google search.

ddavison:
https://github.com/SeleniumHQ/docker-selenium/wiki/Corporate-Proxies

PR #250, opened 2016-07-21
update selenium to 2.53.1, firefox to 47.0.1, chromedriver to 2.23

otsvetkov:
update selenium to 2.53.1, firefox to 47.0.1, chromedriver to 2.23

PR #234, opened 2016-06-08
Update chromedriver to 2.23

smccarthy:
Chromedriver 2.22 is out which supports Chrome 49 - 52.

https://chromedriver.storage.googleapis.com/2.22/notes.txt

PR #226, opened 2016-05-26
Upgrade FF to 46.0.1

Skool:

PR #251, opened 2016-07-22
Add support for Selenium 2.53.1 and Firefox 47.0.1

davehunt:
This adds support for Selenium 2.53.1 and Firefox 47.0.1 in addition to making them the default versions, however I have also introduced an environment variable for each to allow users to opt for previous versions. This will allow us to introduce beta versions that will not be the defaults, and will also offer the ability to workaround compatibility issues between the various dependencies.

This essentially makes #226 and #250 obsolete (except for the ChromeDriver update, but that was also taken care of in #234).

### DIFF
--- a/Base/Dockerfile
+++ b/Base/Dockerfile
@@ -27,7 +27,9 @@ RUN apt-get update -qqy \
 # Selenium
 #==========
 RUN  mkdir -p /opt/selenium \
-  && wget --no-verbose https://selenium-release.storage.googleapis.com/2.53/selenium-server-standalone-2.53.0.jar -O /opt/selenium/selenium-server-standalone.jar
+  && wget --no-verbose https://selenium-release.storage.googleapis.com/2.53/selenium-server-standalone-2.53.0.jar -O /opt/selenium/selenium-server-standalone-2.53.0.jar \
+  && wget --no-verbose https://selenium-release.storage.googleapis.com/2.53/selenium-server-standalone-2.53.1.jar -O /opt/selenium/selenium-server-standalone-2.53.1.jar
+ENV SELENIUM_VERSION=2.53.1
 
 #========================================
 # Add normal user with passwordless sudo
--- a/NodeBase/entry_point.sh
+++ b/NodeBase/entry_point.sh
@@ -33,7 +33,7 @@ fi
 
 SERVERNUM=$(get_server_num)
 xvfb-run -n $SERVERNUM --server-args="-screen 0 $GEOMETRY -ac +extension RANDR" \
-  java ${JAVA_OPTS} -jar /opt/selenium/selenium-server-standalone.jar \
+  java ${JAVA_OPTS} -jar /opt/selenium/selenium-server-standalone-${SELENIUM_VERSION}.jar \
     -role node \
     -hub http://$HUB_PORT_4444_TCP_ADDR:$HUB_PORT_4444_TCP_PORT/grid/register \
     ${REMOTE_HOST_PARAM} \
--- a/NodeBase/entry_point.sh
+++ b/NodeBase/entry_point.sh
@@ -25,6 +25,10 @@ if [ ! -z "$REMOTE_HOST" ]; then
   REMOTE_HOST_PARAM="-remoteHost $REMOTE_HOST"
 fi
 
+if [ ! -z "$FIREFOX_VERSION" ]; then
+  sudo ln -fs /opt/firefox/${FIREFOX_VERSION}/firefox /usr/bin/firefox
+fi
+
 if [ ! -z "$SE_OPTS" ]; then
   echo "appending selenium options: ${SE_OPTS}"
 fi
--- a/NodeFirefox/Dockerfile
+++ b/NodeFirefox/Dockerfile
@@ -6,17 +6,19 @@ USER root
 #=========
 # Firefox
 #=========
-ENV FIREFOX_VERSION 45.0.2
 RUN apt-get update -qqy \
   && apt-get -qqy --no-install-recommends install firefox \
   && rm -rf /var/lib/apt/lists/* \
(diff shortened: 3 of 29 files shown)


davehunt:
@kayabendroth would you mind taking a look at this? I'd like to follow it up with introduction of Selenium 3.0b with GeckoDriver and Marionette support for Firefox 48. I could merge myself, but I'm unfamiliar with the release process.

davehunt:
Perhaps @rubytester could help review/merge/release?

Base/Dockerfile
```diff
@@ -27,7 +27,9 @@ RUN apt-get update -qqy \
 # Selenium
 #==========
 RUN  mkdir -p /opt/selenium \
-  && wget --no-verbose https://selenium-release.storage.googleapis.com/2.53/selenium-server-standalone-2.53.0.jar -O /opt/selenium/selenium-server-standalone.jar
+  && wget --no-verbose https://selenium-release.storage.googleapis.com/2.53/selenium-server-standalone-2.53.0.jar -O /opt/selenium/selenium-server-standalone-2.53.0.jar \
```
kayabendroth: This line has to go away. You download two versions of Selenium.

NodeFirefox/Dockerfile
```diff
@@ -1,22 +1,24 @@
-FROM selenium/node-base:2.53.0
+FROM selenium/node-base:2.53.1
 MAINTAINER Selenium <selenium-developers@googlegroups.com>
 
 USER root
 
 #=========
 # Firefox
 #=========
-ENV FIREFOX_VERSION 45.0.2
```
kayabendroth: Why do you move this line below?

NodeFirefox/Dockerfile
```diff
@@ -1,22 +1,24 @@
...
   && rm -rf /var/lib/apt/lists/* \
-  && wget --no-verbose -O /tmp/firefox.tar.bz2 https://download-installer.cdn.mozilla.net/pub/firefox/releases/$FIREFOX_VERSION/linux-x86_64/en-US/firefox-$FIREFOX_VERSION.tar.bz2 \
-  && apt-get -y purge firefox \
-  && rm -rf /opt/firefox \
-  && tar -C /opt -xjf /tmp/firefox.tar.bz2 \
-  && rm /tmp/firefox.tar.bz2 \
-  && mv /opt/firefox /opt/firefox-$FIREFOX_VERSION \
-  && ln -fs /opt/firefox-$FIREFOX_VERSION/firefox /usr/bin/firefox
+  && apt-get -y purge firefox
+RUN wget --no-verbose -O /tmp/firefox-45.0.2.tar.bz2 https://download-installer.cdn.mozilla.net/pub/firefox/releases/45.0.2/linux-x86_64/en-US/firefox-45.0.2.tar.bz2 \
```
kayabendroth: Why do you install multiple versions of Fx? Do we want that?

NodeFirefox/Dockerfile.txt
```diff
@@ -5,17 +5,19 @@ USER root
 #=========
 # Firefox
 #=========
-ENV FIREFOX_VERSION 45.0.2
```
kayabendroth: Same questions as above.

NodeFirefox/Dockerfile
kayabendroth: If we want to have that, then all of the versions should be defined in an ENV variable, e.g., `FIREFOX_LATEST`, `FIREFOX_MINUS_ONE`, `FIREFOX_MINUS_TWO`. And installing these three versions should be done in a `for`-loop to make the Dockerfile more readable.

Base/Dockerfile
davehunt: This is intentional - to allow the user to control which version is used. We wouldn't want to download many versions, but it would allow us for example to include beta versions.

NodeFirefox/Dockerfile
davehunt: This is the version to be used rather than the version(s) to be downloaded. It made more sense to me to move it to after the downloads, but I can move it back if preferred.

NodeFirefox/Dockerfile
davehunt: I agree that a for loop makes sense here, and if I get a chance to work on this I'll update it. I'm out of the office now for a week, and then traveling for 10 days, so I'm not sure how much time I'll have to work on it. I think the environment variable suggestion is too restrictive as we'll need a way of doing pre-releases too. Perhaps a variable which is a list of versions `FIREFOX_VERSIONS=45.0.2 46.0.1 47.0.1` which can then be used in the for loop?

davehunt:
@kayabendroth would you mind taking another look - I've switch to a for-loop when downloading the Firefox versions now.

NodeBase/entry_point.sh
```diff
@@ -25,6 +25,10 @@ if [ ! -z "$REMOTE_HOST" ]; then
   REMOTE_HOST_PARAM="-remoteHost $REMOTE_HOST"
 fi
 
+if [ ! -z "$FIREFOX_VERSION" ]; then
+  sudo ln -fs /opt/firefox/${FIREFOX_VERSION}/firefox /usr/bin/firefox
+fi
+
```
ddavison: so no matter what time of node you are using, firefox will be in /usr/bin/firefox?  say i'm using NodeChrome, /usr/bin/firefox will exist?

ddavison: i'm aware that it's a no-op if you, of course, do not specify `FIREFOX_VERSION`, but still interesting to hear your reasoning :)

davehunt: Like you point out it will only be created if FIREFOX_VERSION is specified. I'm happy to take another approach if you have an idea?

ddavison: would it make more sense to just put this piece in NodeFirefox and NodeFirefoxDebug?

davehunt: This entry point isn't overridden in some of those, so I think I'd rather keep this here than complicate this further. I'll be opening a new pull request after rebasing from the recent changes shortly.

davehunt:
This has been superseded by #271

PR #279, opened 2016-09-19
Update chromedriver to 2.24

ruffin:
This fixes some bugs and matches vagrant, so 2.24 is the way to go.


ruffin:
whoops, didn't mean to do this on the real one. sorry about that

PR #294, opened 2016-10-04
Update Docker.txt

LarryRuth:
Changed the screen width, height and depth


LarryRuth:
adjust the width

ddavison:
if you need to adjust the size to your liking, you need to build the image yourself 👍 😃

ddavison:
this is way too big.  i can't even see it on my retina display.  i could on the 27" imac, but not on my laptop.  sorry - this is not going into the master branch. if you need to build an image yourself with those display settings, you may do that!

PR #300, opened 2016-10-10
Change FIREFOX_VERSION from ENV to ARG + apt cache clean

F2X:
As FIREFOX_VERSION seems to be used only at build, it should be an ARG to not be confused with env variables. This modification was done for chrome image but not for firefox one.

### DIFF
--- a/Base/Dockerfile
+++ b/Base/Dockerfile
@@ -20,7 +20,7 @@ RUN apt-get update -qqy \
     sudo \
     unzip \
     wget \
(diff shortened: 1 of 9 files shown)


ddavison:
thanks for this!  i did notice i forgot to put that in for firefox.  i'll merge it in shortly

ddavison:
out of curiosity, have you signed the [CLA](https://goo.gl/a2VrTx)?

F2X:
Yes, I've just done it right now

PR #302, opened 2016-10-10
Clear the X server lock in the standalone images

enolan:
This mirrors the fix in #230 into the `entry_point.sh` for the standalone images. `make generate_all` needs to be redone.


ddavison:
regarding the comments of `make generate_all`; regarding the standalone images, it doesn't need to be redone.  you making this change, will deploy this change also to StandaloneChrome and StandaloneFirefox, which fixes this.

Thanks for this!

ddavison:
fyi - i am holding off on putting this in until 3.0.0 stable gets out (should be very soon, and will be released in `3.0.0-actinium`)

Issue #227, opened 2016-05-30
Docker for Mac beta - containers do not start


rnorth:
I think this is potentially, mainly, a Docker for Mac issue, but I'm raising here for visibility and because a fix may be possible on this side.

Here's the issue I've raised in the Docker forums: https://forums.docker.com/t/space-in-new-no-proxy-env-var-breaking-old-containers/14075

In brief:

- Recent versions of Docker for Mac seem to be adding a new environment variable, `no_proxy`, which contains whitespace: `no_proxy=*.local, 169.254/16`
- The entrypoint script for the version of selenium containers I have doesn't properly handle spaces environment variables, causing containers to fail to startup:

```
Waiting xvfb...
-bash: 169.254/16: No such file or directory
Waiting xvfb...
Waiting xvfb...
Waiting xvfb...
Waiting xvfb...
Waiting xvfb...
Waiting xvfb...
Waiting xvfb...
Waiting xvfb...
Waiting xvfb...
```

To see this, with Docker for Mac beta 13 run:
```
$ docker run selenium/standalone-chrome-debug:2.52.0 env

PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
HOSTNAME=0d98e7c76683
no_proxy=*.local, 169.254/16
DEBIAN_FRONTEND=noninteractive
DEBCONF_NONINTERACTIVE_SEEN=true
TZ=US/Pacific
SCREEN_WIDTH=1360
SCREEN_HEIGHT=1020
SCREEN_DEPTH=24
DISPLAY=:99.0
CHROME_DRIVER_VERSION=2.20
LANGUAGE=en_US.UTF-8
LANG=en_US.UTF-8
HOME=/root
```

This line of the entrypoint script seems to need some extra work to deal with spaces in env vars: https://github.com/SeleniumHQ/docker-selenium/blob/master/StandaloneChromeDebug/entry_point.sh#L21

I'll try and submit a patch if I can, but I'd hope the primary fix can be made in the Docker side.

MartinNuc:
I am having the same issue. Unfortunately overriding it with `-e no_proxy=""` doesnt work :(

rnorth:
@MartinNuc actually looking at how I worked around it, I set the `no_proxy` value to `localhost` rather than an empty string. Maybe that makes a difference?

[...]

PR #316, opened 2016-10-19
Handle environment vars. with space properly

stigkj:

If you have this environment variable:

`no_proxy=*.local, 169.254/16`

it is very hard to get the quoting correct. It is easier to just write
the correct entries to the seluser's .bash_aliases file, which will be
included by .bashrc when running xvfb-run as the seluser.

### DIFF
--- a/StandaloneDebug/entry_point.sh
+++ b/StandaloneDebug/entry_point.sh
@@ -18,10 +18,14 @@ fi
 rm -f /tmp/.X*lock
 
 SERVERNUM=$(get_server_num)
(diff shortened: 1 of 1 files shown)


ddavison:
this is a step in the right direction.  I still had to run my node-chrome-debug using `-e no_proxy="" -e HUB_ENV_no_proxy=""`, but that's just because of the Docker issue on OSX

tazle:
This same problem exists in 2.53.1 as reported in #327

a-k-g:
Should solve #227 as well.

Issue #332, opened 2016-11-08
Error forwarding the new session


finspin:
I have the following setup:

- running multiple containers of selenium hub on the same host
- each container running on a different port
- run test A on hub 1
- run test B on hub 2
- Docker version 1.11.0
- docker-compose version 1.8.1

Once the containers are up and running, this is the output of `docker ps`

    17:31:02,500 INFO  - CONTAINER ID IMAGE COMMAND CREATED STATUS PORTS NAMES
    17:31:02,500 INFO  - ab2d99d40c4f selenium/node-chrome-debug:3.0.1-aluminum "/opt/bin/entry_point" 6 seconds ago Up 1 seconds 0.0.0.0:32905->5900/tcp 1478619050087_chromenode_1
    17:31:02,500 INFO  - 875884860b6a selenium/node-firefox-debug:2.53.1-beryllium "/opt/bin/entry_point" 7 seconds ago Up 3 seconds 0.0.0.0:32904->5900/tcp 1478619050087_firefoxnode_1
    17:31:02,500 INFO  - 41cbfd06b201 selenium/hub:3.0.1-aluminum "/opt/bin/entry_point" 11 seconds ago Up 7 seconds 0.0.0.0:32903->4444/tcp 1478619050087_seleniumhub_1
    17:31:02,500 INFO  - ea38707efc36 selenium/node-chrome-debug:3.0.1-aluminum "/opt/bin/entry_point" 15 seconds ago Up 12 seconds 0.0.0.0:32902->5900/tcp 1478619045173_chromenode_1
    17:31:02,500 INFO  - e1877b094cc7 selenium/node-firefox-debug:2.53.1-beryllium "/opt/bin/entry_point" 15 seconds ago Up 13 seconds 0.0.0.0:32901->5900/tcp 1478619045173_firefoxnode_1
    17:31:02,500 INFO  - 18878c35caec selenium/hub:3.0.1-aluminum "/opt/bin/entry_point" 16 seconds ago Up 14 seconds 0.0.0.0:32900->4444/tcp 1478619045173_seleniumhub_1

The problem is that **sometimes** one of the tests fails with the following error. As far as I can tell it's always the test that has been launched as second that fails.

    17:31:10,941 INFO  - [17:31:10] I/hosted - Using the selenium server at http://0.0.0.0:32903/wd/hub
    17:31:10,942 INFO  - [17:31:10] I/launcher - Running 1 instances of WebDriver
[...]

PR #336, opened 2016-11-25
Added support for modifying grid configuration option throwOnCapabili…

finspin:

…tyNotPresent


MicahZoltu:
Is there anything holding up this PR?

I just spent half a day trying to figure out how to make it so I can run tests via `docker-compose` and found that the tests will often run before the nodes are all finished connecting to the hub.  With `throwOnCapabilityNotPresent` set to `true` the hub fails the request from the test driver setup because it doesn't have any capabilities yet.  I believe setting this would resolve the problem and make it so the requests from my tests will queue up until the necessary node is fully connected.

finspin:
@MicahZoltu yes, this PR fixes this exact issue you described. Until it's merged, you can use a workaround described in https://github.com/SeleniumHQ/docker-selenium/issues/332.

MicahZoltu:
I ended up solving the problem myself by building my own docker image.  I would have preferred to use an environment variable as it would have been a much cleaner solution, but this got things working for now.
https://github.com/Zoltu/bags-client/blob/master/functional-tests/Dockerfile-selenium-hub

PR #339, opened 2016-12-02
add chinese font in base contaimer

b4456609:
The screenshots in selenium chrome node shows that Chinese characters display with square boxes. Install Chinese font could solve the problem.

### DIFF
--- a/Base/Dockerfile
+++ b/Base/Dockerfile
@@ -20,6 +20,7 @@ RUN apt-get update -qqy \
     sudo \
     unzip \
     wget \
+    fonts-noto-cjk \
   && rm -rf /var/lib/apt/lists/* /var/cache/apt/* \
   && sed -i 's/securerandom\.source=file:\/dev\/random/securerandom\.source=file:\/dev\/urandom/' ./usr/lib/jvm/java-8-openjdk-amd64/jre/lib/security/java.security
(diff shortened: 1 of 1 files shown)


elgalu:
Note installing `fonts-noto-cjk` will add 126.7 MB to the docker image, perhaps is better generating a separate image in this case.

HackToday:
Adding Chinese support is good. however not sure if those containers could accept such locale support for general cases. As different language may have different requrements

ddavison:
i'm going to keep this PR open just for reference, but i've assigned a milestone that i would like to achieve with these containers, and that is internationalization.

diemol:
Someone sent a PR to https://github.com/elgalu/docker-selenium/pull/153, with the same purpose. 
The main difference is that the installed packages only need 30MB more. Maybe this is an alternative?

Cyperwu:
Is this Issue going any further?

ddavison:
regarding your comment, @diemol, i tried a before-after using these packages:

- fonts-ipafont-gothic
- fonts-wqy-zenhei

Before and after had the same result - and the result looked like this:
screen shot 2018-01-22 at 1 59 24 pm

diemol:
Closing this one since we merged #676

PR #346, opened 2016-12-14
Allow X11 forwarding

BrennanRoberts:

Running the browser inside xvfb by default prevents users from
forwarding their hosts' X socket to the container. When attempting to
run:

`docker run -e -e DISPLAY=$DISPLAY -v /tmp/.X11-unix:/tmp/.X11-unix selenium/standalone-chrome`

Here, we only run inside xvfb after checking whether there is not an
existing X server. Adds the `x11-xserver-utils` package to the NodeBase
image, used for that query.

### DIFF
--- a/NodeBase/Dockerfile
+++ b/NodeBase/Dockerfile
@@ -22,6 +22,7 @@ RUN echo "${TZ}" > /etc/timezone \
 RUN apt-get update -qqy \
   && apt-get -qqy install \
     xvfb \
+    x11-xserver-utils \
   && rm -rf /var/lib/apt/lists/* /var/cache/apt/*
 
 #==============================
--- a/StandaloneChrome/entry_point.sh
+++ b/StandaloneChrome/entry_point.sh
@@ -17,9 +17,15 @@ SERVERNUM=$(get_server_num)
 
 rm -f /tmp/.X*lock
 
-xvfb-run -n $SERVERNUM --server-args="-screen 0 $GEOMETRY -ac +extension RANDR" \
+# Use xvfb only if there is not an x server running already
+if xset q &>/dev/null; then
   java ${JAVA_OPTS} -jar /opt/selenium/selenium-server-standalone.jar \
-  ${SE_OPTS} &
+    ${SE_OPTS} &
+else
+  xvfb-run -n $SERVERNUM --server-args="-screen 0 $GEOMETRY -ac +extension RANDR" \
+    java ${JAVA_OPTS} -jar /opt/selenium/selenium-server-standalone.jar \
+    ${SE_OPTS} &
+fi
 NODE_PID=$!
 
 trap shutdown SIGTERM SIGINT
--- a/StandaloneFirefox/entry_point.sh
+++ b/StandaloneFirefox/entry_point.sh
@@ -17,9 +17,15 @@ SERVERNUM=$(get_server_num)
 
 rm -f /tmp/.X*lock
 
-xvfb-run -n $SERVERNUM --server-args="-screen 0 $GEOMETRY -ac +extension RANDR" \
+# Use xvfb only if there is not an x server running already
+if xset q &>/dev/null; then
   java ${JAVA_OPTS} -jar /opt/selenium/selenium-server-standalone.jar \
-  ${SE_OPTS} &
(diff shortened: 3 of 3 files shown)


BrennanRoberts:
This patch is intended to address #2.

ddavison:
i'm not sure you intended to change the standalone containers.  these standalone containers don't/won't need VNC. that's only the node-debug containers

BrennanRoberts:
Hi @ddavison, we use the standalone containers for our tests using the workflow:
1. Start a standalone selenium container
1. Start a test container and point it to the selenium container
Currently, in order to view the browser when running these tests locally, we need to change the standalone image to the "-debug" variant, and then use VNC. This patch would allow us to, instead, start the standalone container with the X forwarding, and the browser would render on the host automatically without the need to change the image or start a VNC viewer.

Is there another way to achieve this result?

NodeBase/Dockerfile
```diff
@@ -22,6 +22,7 @@ RUN echo "${TZ}" > /etc/timezone \
 RUN apt-get update -qqy \
   && apt-get -qqy install \
     xvfb \
+    x11-xserver-utils \
```
BrennanRoberts: This package provides `xset`. It's already on the Chrome images as a dependency, this just adds it to the other ones.

BrennanRoberts:
Looking for more feedback on other use cases this should support.

HackToday:
Interesting to this PR
1> Right now, we have a selenium container running, and our tests (if not run locally, as locally can have Firefox UI, so local test is not problem) run remotely against that container.  The standalone container not have VNC, so I can not view the browser. If I want to view the actual browser in the container, I need to use Debug selenium container.

 So did your x-forward ways allow view the browser for standalone container ?
And how can I view the browser if not use VNC client ?

BrennanRoberts:
@HackToday I'm not sure this patch will make it any easier for you to view the browser when the selenium container is running on a different host. The case it does improve is where you want to run a standalone container locally using the 'remote' driver, and want to see it without using VNC.

HackToday:
> The case it does improve is where you want to run a standalone container locally using the 'remote' driver, and want to see it without using VNC

OK. Thanks. If that case, does it mean I still need to make sure the local host need have GUI environment. or else how can we view such browser ?

BrennanRoberts:
>  If that case, does it mean I still need to make sure the local host need have GUI environment. 

I believe you could use remote X forwarding for this case.

lucasvanlierop:
@BrennanRoberts Do you have an example of how you use this? For example how would you get a live view of what Selenium is running in Chrome?

I already know Chrome works, verified this taking the following steps:
- Configure a selenium service for Docker Compose,
- Run it as my own user (since x cannot run as root)
- Give it a home directory chrome can write to
- Start the selenium container: `docker-compose up selenium-chrome
- Run  Chrome: `docker-compose exec selenium-chrome google-chrome`

My Docker Compose config is:
```yml
 selenium-chrome:
    image: selenium/standalone-chrome
    user: {MY_USER_ID}:{MY_GROUP_ID}
    environment:
      DISPLAY: $DISPLAY
      HOME: /home
    volumes:
      - ./docker/selenium-chrome/entry_point-x11.sh:/opt/bin/entry_point.sh:ro
      - /tmp/.X11-unix:/tmp/.X11-unix
      - /tmp/home/:/home/
```

lucasvanlierop:
@BrennanRoberts As always my error was something completely unrelated. I misconfigured my BDD tool to use `http://selenium-chrome:4444` instead of `http://selenium-chrome:4444/wd/hub` which resulted in selenium giving an error response instead of starting chrome. This error did not appear anywhere of course (until I started debugging). Now my

lucasvanlierop:
I've been using this patch for a while now and it works really well

bryanlarsen:
Another use case: so that the GPU can be used for tests.    We can either hook up to a running X server, or hook up via compose to a separate docker image running Xdummy.

bryanlarsen:
I ended up using https://github.com/plumbee/nvidia-virtualgl-selenium-node-chrome instead.

diemol:
Closing PR since no one replied in the related #2 issue.

PR #348, opened 2016-12-15
Moving port assignment to environment variable to allow for assignment

wesmcouch:

This PR will allow the selenium nodes to be run from marathon/mesos. Before you were unable to assign the port at the proper time when the container was being built. This way allows the dynamically assigned port from mesos to be assigned to the node container.

This has been tested in marathon and I am able to scale the nodes as needed when using a docker image built with this configuration. An example marathon definition file can be provided to show how this works if needed.

This is a better way of doing the fix described in this PR that is currently open:
https://github.com/SeleniumHQ/docker-selenium/pull/104

### DIFF
--- a/NodeChrome/Dockerfile
+++ b/NodeChrome/Dockerfile
@@ -44,6 +44,7 @@ RUN wget --no-verbose -O /tmp/chromedriver_linux64.zip https://chromedriver.stor
 ENV NODE_MAX_INSTANCES 1
 ENV NODE_MAX_SESSION 1
 ENV NODE_REGISTER_CYCLE 5000
+ENV NODE_PORT 5555
 COPY generate_config /opt/selenium/generate_config
 RUN chmod +x /opt/selenium/generate_config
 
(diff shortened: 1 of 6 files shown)


HackToday:
Hi @wesmcouch  I am not sure if I understand this PR. 
do you mean marathon and mesos can not run selenium containers ? (if without this PR)

I think marathon allow port mapping when run docker containers, so not know what's the issue about.

wesmcouch:
Hey @HackToday,

Mesos can run the selenium hub just fine, but when running the nodes it becomes an issue. 

There issue is using SE_OPTS to tell the node what port marathon has assigned to the container is too late during the process of creating the container. For some reason marathon is unable to link this assigned port and port 5555 during this process. The node will be able to connect to the hub but you are unable to connect to the node on the assigned port.

Using the fix described in the PR, you are able to put the assigned port in as an environment variable  using HOST mode in marathon and it will connect just fine.

Here is my current configuration that works really well with the above PR. I can spin up any number of nodes with this configuration.

image

image

HackToday:
> **The issue is using SE_OPTS to tell the node what port marathon has assigned to the container is too late during the process of creating the container. For some reason marathon is unable to link this assigned port and port 5555 during this process. The node will be able to connect to the hub but you are unable to connect to the node on the assigned port.**

Not quite understand here. As hub started in host mode,(it has host ip, and port) and then start node to register with such known ip and port.

Node could connect to hub, and you can also connect to node.

ddavison:
i don't have a problem moving this in. parameterizing things just makes our images more versatile. next time, please don't touch "Dockerfile" files though - only edit the Dockerfile.txt files

wesmcouch:
Sounds good, thanks!

ddavison:
thanks for this! 👍

PR #355, opened 2016-12-22
add “version” property to capabilities object in generate_config

mirkotschaeni:
I think that it would make sense to add a "version" property to the capabilites object in generate_config.
The version property is a standard capability key and is helpful for instance in grid setups with multiple versions of the same browser type (version based capability matching).


bhaku:
I also need this. Now you must login to container to check which version of browser is used.

ddavison:
merged in, thanks!  great fix

PR #366, opened 2017-01-13
Upgrade to firefox 50.1.0, geckodriver v0.13.0

rjatkins:


ddavison:
thanks!  don't worry about the failing CI.  1 in every 10 builds or so it fails when it shouldn't.

AkeemMcLennon:
I noticed that the Dockerfile is autogenerated from `generate.sh`, but I'm curious when the script actually gets run. Was its output supposed to be part of this commit?

PR #372, opened 2017-01-27
Made it possible to build customers version of these docker images.

maf23:
I tried to build private copies of these images and manually set
NAME and VERSION when launching make. Unfortunately it turned out
that the NAME did not propagate so for example NodeFirefox wanted to
build on selenium/node-base:MY_VERSION which failed since that didn't
exist. This patch propagates NAME to the generate.sh-scripts so the
images build on each other.

### DIFF
--- a/Hub/generate.sh
+++ b/Hub/generate.sh
@@ -1,9 +1,10 @@
 #!/bin/bash
 VERSION=$1
+NAME=$2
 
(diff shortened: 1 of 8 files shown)


ddavison:
these are our makefiles, and we are going to leave our name on them.  the beauty of what we have here, is that it's open source, so you can keep "your" makefile and use it for your customers when building your own images; but this PR is not going in.  sorry!

ddavison:
you *can*  fork this repository and maintain these images yourself, as well! :+1:

Issue #390, opened 2017-02-15
Timeout seems to be too large


wheleph:
## Meta -
Image(s):  
hub

## Actual Behavior -
The timeout value is 30000 (https://github.com/SeleniumHQ/docker-selenium/blob/master/Hub/Dockerfile.txt#L14). But Selenium assumes that the value is in seconds: https://github.com/SeleniumHQ/selenium/blob/b1c2cf6ceeab96856cf42feb4e84d2400a0a61eb/java/server/src/org/openqa/grid/internal/utils/configuration/StandaloneConfiguration.java#L194

As a result the hub waits for more than 8 hours before ditching stale client sessions. It might be better to set it to something lower like 30.

ddavison:
good catch, would you want to try your hand at a pull request?

PR #394, opened 2017-02-16
Set proper GRID_TIMEOUT

wheleph:


wheleph:
Fixes #390

Hub/Dockerfile
```diff
@@ -16,7 +16,7 @@ ENV GRID_NEW_SESSION_WAIT_TIMEOUT -1
 ENV GRID_JETTY_MAX_THREADS -1
 ENV GRID_NODE_POLLING  5000
 ENV GRID_CLEAN_UP_CYCLE 5000
-ENV GRID_TIMEOUT 30000
+ENV GRID_TIMEOUT 30
```
conversica-aaronpa: Too low, now I have to implement an override

ddavison: that's ok :)  that's how they're meant to be used.  since this is an environment variable passed to the container, you can set it to whatever you need it to be.  it would be hard for this default value to match "everybody's" standards.

PR #400, opened 2017-02-23
Consolidated image name(space) and maintainer and updated Makefile.

camiller2:


ddavison:
i've denied a change like this in the past since these are our instances, and you are very well able to create your own.

that being said - i'm very open minded.  whey did you choose to make this change? what is your reasoning? 

Also, you deleted the template that came with the pull request.  please tell me whether you have signed the CLA

camiller2:
I'm not sure what the CLA is (so I probably haven't signed it :-p).  There are two reasons for the change:

1) DRY
2) I forked the repo to get a version of 2x with a particular version of firefox; adding this change made it easy to update the namespace in all the generated Dockerfile files.

ddavison:
cla was in the template that you deleted 🤐 

https://github.com/SeleniumHQ/docker-selenium/blob/master/.github/PULL_REQUEST_TEMPLATE.md

alright, let me do some investigation to see if this is a standard practice. if it is, i wouldn't mind putting this in. thanks for the justification

jalmansor:
I have actually been wanting to put in a change like this as well, since we have a need to generate our own images on a regular basis, and this would make it easier to push to our own repo without breaking your existing build workflow.

ddavison:
ok, i'd like to give this PR another chance to get put in. 

some things need to happen before it's merged:

1. We have to make sure "all" nodes are covered with this change.  (there is a phantomjs node now which will need to be covered)
2. it turns out `MAINTAINER` is [deprecated](https://docs.docker.com/engine/reference/builder/#maintainer-deprecated). Instead, we should try using `LABEL maintainer "..."`

jalmansor:
@camiller2 we got their attention! Let's get this fixed up and ready for them to merge!

ddavison:
this was put in with 9e2d61099c51aff0c89335ce0b198d821080a16d

PR #414, opened 2017-03-14
Update images to selenium 3.3.1 and geckodriver 0.15.0

BeyondEvil:

### DIFF
--- a/Base/Dockerfile
+++ b/Base/Dockerfile
@@ -34,8 +34,10 @@ RUN apt-get update -qqy \
 #==========
 # Selenium
 #==========
+ARG SELENIUM_VERSION=3.3.1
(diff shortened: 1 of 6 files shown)


ddavison:
I appreciate the effort, but this goes against the release versioning i've setup for this.  However, i "do" like the idea about adding the build ARG's to specify the version!  nice job.

I won't be merging this however, again - as it goes against the versioning.  but don't let that deter you from more pr's!  could you by chance recreate this PR with just the FF changes?

BeyondEvil:
@ddavison Not sure I understand about the versioning... but I'll make a new PR with the ARGS and we can take it from there.

What about the hardcoded versions in the tests? I can't see the rationale for that.

PR #417, opened 2017-03-17
feat(encrypt-vnc): add encrypted WebSocket proxy

0xC4N1:
This commit introduces a WebSocket proxy on port 6900 for encrypted VNC
sessions via WebSocket clients like noVNC.

Usage: See README.md

- [ X ] By placing an `X` in the preceding checkbox, I verify that I have signed the [Contributor License Agreement](https://github.com/SeleniumHQ/docker-selenium/blob/master/CONTRIBUTING.md#contributing-code-to-selenium)

### DIFF
--- a/NodeChromeDebug/Dockerfile
+++ b/NodeChromeDebug/Dockerfile
@@ -17,6 +17,14 @@ RUN apt-get update -qqy \
   && mkdir -p /root/.vnc \
   && x11vnc -storepasswd secret ~/.vnc/passwd
 
+#============
+# Websockify
+#============
+RUN apt-get update -qqy \
+  && apt-get -qqy install \
+    websockify python-pkg-resources \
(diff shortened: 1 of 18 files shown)


ddavison:
i would hate to add this to all of the base images, for the fact that most people probably won't be using this, yet we would make everybody install it (so it takes their time) even if they are not using it.

I really have to find out a better to way to manage this sort of stuff.  if I merged this in, it would probably be 1 of 2 cases.

1) In the dockerfile, we have some sort of conditional, where if they want it it will install (this could be used for many things.
2) I release completely new images with this in it.  For example: selenium/node-chrome-debug-encrypted, selenium/node-firefox-debug-encrypted.

number (1) is certainly preferred, as there are already a lot of images that are there.

diemol:
I'll close this PR since there was no reaction after the 2nd comment. Even though this could be helpful, this would add more packages to an already big image, and not many users would use it.

Issue #415, opened 2017-03-14
Newer versions of Firefox have dropped some libraries ; breaks older versions


awilkins:
## Meta -
Image(s):  Firefox

Docker Version:  17-03, 3.0.1-germanium

OS: 
Ubuntu 16.04

## Expected Behavior -

* Set FIREFOX_VERSION to 45.0.2
* Build images
* Testing on the image works

## Actual Behavior -

* XPCOMGlueLoad error

Shelling the docker container and checking library links reveals...

```
# Removed libraries that were actually found

seluser@eaf279429169:/opt/firefox-45.0.1$ ldd libxul.so 
	libmozsqlite3.so => not found
	liblgpllibs.so => not found
	libasound.so.2 => not found

```

Theory : Firefox 52 (the version you get now if you `apt-get install firefox` on Ubuntu) has dropped dependencies on these libraries.

Therefore the Dockerfile behaviour of installing the latest version of Firefox to get the dependencies isn't sound if you're not building the latest version.

Possible solutions : 
- Have the script read through the apt cache and find the available version that's closest to the version you specified in FIREFOX_VERSION, and install that instead.

On Ubuntu 16.04 (the base of the image) this is limited to versions 45.0.2 and 52 at present. Maybe work out which version of Firefox dropped these libraries and pick one either side of that divide.

Happily I wanted Firefox 45 (my target user base is trapped in a government-sponsored time warp).

What I actually did :
- Patched the Dockerfile for the base image to specifically install Firefox 45.0.2+build1-0ubuntu1

awilkins:
Success : built an image that runs Firefox 45 properly.

gangadharahn:
Hi,

I am also trying to build firefox images for lower versions, could please tell me the changes you did on base?

Thanks for your help in advance.

Regards,
Ganga

awilkins:
The change I did was to look up the available package versions of Firefox for the base 16.04 version of Ubuntu that's used for the Docker base image, and change the Dockerfile template to specifically install that version instead of latest.

e.g. 

```
&& apt-get -qqy --no-install-recommends install firefox=45.0.2+build1-0ubuntu1 \
```

Given there's a FIREFOX_VERSION environment variable, I guess it should detect versions under 52 and make this change to the file - or define another variable like FIREFOX_DEPENDENCY_PACKAGE_VERSION and set that to latest or that version 45 depending.

ddavison:
possibly related to #418

[...]

Issue #397, opened 2017-02-22
Firefox Browser is Crashing frequently for both latest and previous version


oliyavan:
## Meta -
Image(s):  selenium/node-firefox
Docker Version:  1.12.6
OS: RHEL 7
## Expected Behavior -
we are running selenium test scripts for our internal application testing, we start looking to feasibility of run nodes in the docker containers instead of VMs, we configured and we running the test cases in normal case everything working, we try to execute series it should give the result with a screenshrtst.
## Actual Behavior -
Instead of that browser tab is crashing again and again, we tried with old version of firefox also but still, 
browser tab getting crashed.
![Uploading firefox_crashing.png…]()

So we diced to check with Debug image, please find the attachment for screen shot.

Error logs of the node:
```
For documentation on this error, please visit: http://seleniumhq.org/exceptions/no_such_element.html
Build info: version: '3.0.1', revision: '1969d75', time: '2016-10-18 09:48:19 -0700'
System info: host: '83f5c427733e', ip: '172.17.0.4', os.name: 'Linux', os.arch: 'amd64', os.version: '3.10.0-327.22.2.el7.x86_64', java.version: '1.8.0_121'
Driver info: org.openqa.selenium.firefox.FirefoxDriver
[...]

PR #418, opened 2017-03-21
Firefox crash for version lower than 52

jackTheRipper:
Hi,

This PR fix the following issues with Firefox:
 - https://github.com/SeleniumHQ/docker-selenium/issues/397
 - https://github.com/SeleniumHQ/docker-selenium/issues/415

The fix adds the required dependencies for the Firefox versions lower than 52 and make it stable again.

- [ X ] By placing an `X` in the preceding checkbox, I verify that I have signed the [Contributor License Agreement](https://github.com/SeleniumHQ/docker-selenium/blob/master/CONTRIBUTING.md#contributing-code-to-selenium)


ddavison:
will updating the version to 52 fix this?  if so, then [we're all good now](https://github.com/SeleniumHQ/docker-selenium/blob/master/NodeFirefox/Dockerfile#L13)

jackTheRipper:
Updating to the 52 doesn't solve issue. I want to run my tests against different version of Firefox.

Just updating the version without that `if` wouldn't the issue. 

If you take a look at both of the issues I mentioned the problem comes from some missing dependencies.

I have tested both version on my side (52, 51.0.1). Without the packages I added to the image the 51.0.1 crash. The 52 as you say doesn't have the issue. 

I am using currently the version '3.1.0-astatine'.

Maybe the PR shouldn't go to master but to that specific branch?

jackTheRipper:
@ddavison question unrelated to my PR: instead of having a Selenium base installing both Java dependencies and Selenium, wouldn't it be easier to have a base Docker image which setups Java and a Docker image which setups Selenium?

jackTheRipper:
I performed some tests available in this [gist](https://gist.github.com/jackTheRipper/af09b57f7308b6f0c9983ed19615405c)

It compares the current image without my modifications to the one with my modifications for the version 51.0.1.

jackTheRipper:
@ddavison any news regarding the PR?

brandomr:
@jackTheRipper still getting weird timeouts/tab crashes even using this PR

jackTheRipper:
@bandomr: are you using http basic authentication for your tests?

brandomr:
@jackTheRipper no, I'm just running grid/node-firefox against some sites that don't require auth

jackTheRipper:
@brandomr: could you provide some logs?

brandomr:
I'm seeing:

```
[Parent 160] WARNING: pipe error (56): Connection reset by peer: file /builds/slave/m-rel-l64-00000000000000000000/build/src/ipc/chromium/src/chrome/common/ipc_channel_posix.cc, line 322
[Parent 160] WARNING: pipe error (49): Connection reset by peer: file /builds/slave/m-rel-l64-00000000000000000000/build/src/ipc/chromium/src/chrome/common/ipc_channel_posix.cc, line 322
[Parent 160] WARNING: pipe error (46): Connection reset by peer: file /builds/slave/m-rel-l64-00000000000000000000/build/src/ipc/chromium/src/chrome/common/ipc_channel_posix.cc, line 322

###!!! [Parent][MessageChannel] Error: (msgtype=0x2C0085,name=PBrowser::Msg_Destroy) Channel error: cannot send/recv
```

and then:
```
02:06:31.889 INFO - Session f47c7085-5135-4792-9714-874cf9147084 deleted due to client timeout
1491357991893	Marionette	INFO	New connections will no longer be accepted
```

jackTheRipper:
Yes I had the same messages.

@brandomr could you re use the example I shared in my gist and provide the whole logs? 

I need to see which capabilities you are using. 

Are you running the tests  against Firefox 51.0.1 or lower?

brandomr:
@jackTheRipper no I was using the default (FF 52+). I'll downgrade and try to get you full logs. Thanks for the help with this stuff

jackTheRipper:
@brandomr after re reading your comments it's not related to my PR. 
You are talking about a version of Firefox up to 52. My PR is here for fixing an issue for the version lower than 52.

In your case I think that your issue is related to the geckodriver

jackTheRipper:
@ddavison What is the status about the pull-request?

ddavison:
well, can't do much until the conflicts have been resolved :)  but i'd like some other people's opinion on the matter.  

@beyondevil ?
@tparikka ?

jackTheRipper:
@ddavison sorry for not fixing it earlier. Here it is.

BeyondEvil:
Some quick thoughts.

General:
If you change the FF version (ie. rebuild the image(s)) you can add the missing library (libasound 2) at that point. As we kind of expect at that point that you know what you're doing.

Specific:
Not sure you need anything else than the libasound 2 library, as I fail to see how this solves the other issue (instability). If that is the case, skip the `if` and just install libasound 2 regardless as there are a number of "unused" libraries installed anyway. libasound 2 is fairly lightweight and won't bloat the image by much. But I could definitely be in the wrong here. Interesting to see if others with this issue finds the solution in this PR.

My 0.2 anyway. 
@ddavison 
This is kind of what we talked about with trying to make the images more versatile and not locked to specific combinations of browser/driver/selenium versions. I should have some time next week (not this coming one) to start looking at a solution for that. But I want us to discuss testing first. Maybe set up a meeting for the coming week?

navels:
bump

diemol:
Hi all,

I agree with @BeyondEvil's last comment, but the main question for me is if this was still an issue... 
I am not sure how many people are still testing with a Firefox version lower than 52. I am sure there are, but I guess this is not the common case. 
What do you think?

diemol:
I'll close this PR since we didn't get more feedback. We can take a look at it in case it makes sense to treat this case in the images.

PR #419, opened 2017-03-22
Removed copy command from debug image

ersuniljoshi:

### DIFF
--- a/StandaloneChromeDebug/Dockerfile
+++ b/StandaloneChromeDebug/Dockerfile
@@ -53,7 +53,6 @@ RUN apt-get update -qqy \
 #==============================
 # Scripts to run Selenium Node
 #==============================
-COPY entry_point.sh /opt/bin/entry_point.sh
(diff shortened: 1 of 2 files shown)


ddavison:
what's the reason for this change?  this piece is needed.  additionally - in the line below the one you took out, there would be a failure since you can't chmod a file that doesn't exist ;)

ersuniljoshi:
@ddavison the reason I have removed this file is, its already being copied in the above docker image(i.e. StandaloneFireFox) and this copy will fail as "entry_point.sh" doesnot even exists in the folder named "StandaloneFireFoxDebug". After removing this file I was able to build the debug image. Otherwise its fails with Error "No such file or directory"

ddavison:
hey! good call - let me take a closer look

diemol:
@ersuniljoshi but this is needed, even if the base image has an entry_point.sh, those Dockerfiles overwrite it because the debug one has VNC running.

This change does not apply. I'll close this PR.

Issue #465, opened 2017-04-22
Chrome sessions hang randomly with node-chrome due to DBUS issue


testphreak:
## Meta -
Image(s):
node-chrome. May affect other chrome and firefox images as well.

Image Version:
3.3.1-cesium

Docker version:
17.04.0-ce, build 4845c56

OS:
Ubuntu 16.04.2 LTS

## Expected Behavior -
Tests execute as usual without hanging.

## Actual Behavior -
@diemol, @ddavison spent a good deal of time last night trying to get to the bottom of the issue. It's related to this change https://github.com/SeleniumHQ/docker-selenium/commit/88584def77585863a226ca72bbce0f9dce561242 made by @a-k-g to resolve #358. 

Due to this change, none of the node-chrome images after 3.0.1 are working for me and the timeline of the commit confirms that. Would it be possible to add a fix that would work for all images - node-chrome, node-chrome-debug, standalone-chrome and standalone-chrome-debug images?

I passed in `DBUS_SESSION_BUS_ADDRESS=/dev/null` as an environment variable to node-chrome when starting the containers and all my tests now execute without random hangs with 3.4.0-actinium. It would be better if this variable is set in the image rather than us having to pass it in every time. 

I also found out that I am not alone in this issue and several including @pwaller and others have run into this issue recently https://github.com/SeleniumHQ/docker-selenium/issues/87#issuecomment-284238845.

diemol:
I was able to reproduce it with the Chrome images from the 3.4.0-bismuth release. Summarising, under this use case (open and closing Chrome many times, at least 50):

* node-chrome -> Does not work
* node-chrome-debug -> Works
* standalone-chrome -> Does not work
* standalone-chrome-debug -> Does not work

I am building locally the Dockerfiles now with some changes to see if it works. I'll keep you updated.

testphreak:
Awesome! Thanks @diemol!

diemol:
@testphreak Firefox images seem to work fine. By fine I mean that the test does not hang, it either fails or passes (I am just clarifying it because I am not 100% sure if the latest version of Geckodriver - 0.16 and Firefox 53 are working well, so nothing related to the DBUS_SESSION_BUS_ADDRESS).

diemol:
@testphreak, I just created a PR for this. Please check it, any help with checking it with your local tests would be great.

testphreak:
@diemol, your change looks good. Verified your change locally. Thanks for reverting the change back to #182.

[...]

Issue #358, opened 2016-12-27
standalone-chrome-debug hangs becouse DBUS_SESSION_BUS_ADDRESS is lost with sudo


Sovetnikov:
When selenium is started via "sudo -E -i seluser" on -debug standalone, DBUS_SESSION_BUS_ADDRESS=/dev/null variable is lost and chrome hangs periodically on startup.

Just test it:
root:# docker run --rm selenium/standalone-chrome-debug env | grep DBUS
DBUS_SESSION_BUS_ADDRESS=/dev/null
root:# docker run --rm selenium/standalone-chrome-debug sudo -E -i -u seluser env | grep DBUS

Please fix this in entry_point.sh by adding "sudo -E -i seluser DBUS_SESSION_BUS_ADDRESS=/dev/null"
It's very annoying bug.

https://github.com/SeleniumHQ/docker-selenium/blob/aabd14c0b8bebdef1e492ffde927b7815ee2b19a/StandaloneDebug/entry_point.sh

a-k-g:
+1 for this issue. I was hitting #87 repeatedly even after the fix seemed to have been merged. Thanks @Sovetnikov for finding the root cause. I've made an attempt to fix this issue in #381.

Sovetnikov:
6 months later this issue still exists

diemol:
How can we reproduce it with the latest images @Sovetnikov?

Can you please provide a script that we can use?

Sovetnikov:
As in initial issue description:
docker run --rm selenium/standalone-chrome-debug sudo -E -i -u seluser env | grep DBUS
or
docker run --rm selenium/standalone-chrome:3.4.0-dysprosium sudo -E -i -u seluser env | grep DBUS

Two weeks ago i made update to 3.4.0-dysprosium image and again getting timeouts on new selenium sessions.

Sovetnikov:
3.1.0-astatine works fine:
docker run --rm selenium/standalone-chrome:3.1.0-astatine sudo -E -i -u seluser env | grep DBUS
seems DBUS fix is lost somewhere between 3.1 and 3.4

Sovetnikov:
Two lines in custom Dockerfile fixed it for my:
RUN echo 'Defaults env_keep += "DBUS_SESSION_BUS_ADDRESS"' >> /etc/sudoers
RUN echo "DBUS_SESSION_BUS_ADDRESS=/dev/null" >> /etc/environment

diemol:
[...]

PR #469, opened 2017-04-23
Fixing #87 (again), #465 and similar ones

### LINKED ISSUE #87: Chromedriver frequently hangs when attempting to start a new session.
Using the standalone-chrome or node-chrome images, I often observe a timeout when attempting to launch a new Chrome session.

Steps to reproduce:

1) launch a standalone chrome instance: `docker run -d --name chrome selenium/standalone-chrome:2.47.1`
2) build the test container: `docker build -t selenium/test:local ./Test`
3) run the repeat test script: `./test-repeat.sh chrome`
4) If necessary, repeat steps 1 and 3 until the problem is observed

Note:

In order to narrow the focus of the test to launching new sessions, you may want to temporarily modify `Test/smoke-test.js` to omit the part of the test that tries navigating to github.com after starting the session. After modifying the script, make sure to rebuild the selenium/test:local image to pick up the changes.

Expected results:

All 50 sessions launch and quit successfully - the test passes.

Actual results:

One of the session launch commands will hang indefinitely.

Docker host: boot2docker v1.7.0 (Tiny Core Linux)

diemol:
Originally #87 was solved by #182 by adding `ENV DBUS_SESSION_BUS_ADDRESS=/dev/null` to the Chrome and Firefox Dockerfiles. 

Nevertheless #358 was raised and due to that the command changed via #386 to `RUN echo "DBUS_SESSION_BUS_ADDRESS=/dev/null" >> /etc/environment`. 

But after that, #87 started showing up again.

So this PR leaves the Dockerfiles in the same state as the commit made in #182 that fixed #87.

I tested the locally generated Chrome images it with the current tests but also with a simple test that help me to reproduce #465 and #87. Here is the [simple test](https://gist.github.com/diemol/df6f97b35c3943358a78d8200bbeae35)

I was not able to reproduce the error reported in #358, maybe @Sovetnikov or @a-k-g can provide a test case to reproduce it before this PR is considered to be merged?

@testphreak, maybe you can also generate the images locally based on this PR and test it?

### DIFF
--- a/NodeChrome/Dockerfile.txt
+++ b/NodeChrome/Dockerfile.txt
@@ -57,15 +57,14 @@ COPY generate_config /opt/selenium/generate_config
 RUN chmod +x /opt/selenium/generate_config
 
 #=================================
-# Chrome Launch Script Modication
+# Chrome Launch Script Modification
 #=================================
 COPY chrome_launcher.sh /opt/google/chrome/google-chrome
 RUN chmod +x /opt/google/chrome/google-chrome
 
 RUN chown -R seluser:seluser /opt/selenium
 
-# Following line fixes
-# https://github.com/SeleniumHQ/docker-selenium/issues/87
-RUN echo "DBUS_SESSION_BUS_ADDRESS=/dev/null" >> /etc/environment
+# Following line fixes https://github.com/SeleniumHQ/docker-selenium/issues/87
+ENV DBUS_SESSION_BUS_ADDRESS=/dev/null
 
 USER seluser
--- a/Makefile
+++ b/Makefile
@@ -227,10 +227,10 @@ release: tag_major_minor
 	docker push $(NAME)/standalone-firefox-debug:$(MAJOR_MINOR_PATCH)
 
 test:
-	./test.sh
-	./sa-test.sh
-	./test.sh debug
-	./sa-test.sh debug
+	VERSION=$(VERSION) ./test.sh
+	VERSION=$(VERSION) ./sa-test.sh
+	VERSION=$(VERSION) ./test.sh debug
+	VERSION=$(VERSION) ./sa-test.sh debug
 
 .PHONY: \
 	all \
--- a/NodeBase/Dockerfile.txt
+++ b/NodeBase/Dockerfile.txt
@@ -10,7 +10,6 @@ RUN apt-get update -qqy \
   && apt-get -qqy install \
     locales \
     xvfb \
-    dbus \
   && rm -rf /var/lib/apt/lists/* /var/cache/apt/*
 
 #==============================
--- a/NodeFirefox/Dockerfile.txt
+++ b/NodeFirefox/Dockerfile.txt
@@ -53,8 +53,7 @@ COPY generate_config /opt/selenium/generate_config
 RUN chmod +x /opt/selenium/generate_config \
   && chown -R seluser:seluser /opt/selenium
 
-# Following line fixes
-# https://github.com/SeleniumHQ/docker-selenium/issues/87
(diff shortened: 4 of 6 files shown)


NodeChrome/Dockerfile.txt
```diff
@@ -57,15 +57,14 @@ COPY generate_config /opt/selenium/generate_config
 RUN chmod +x /opt/selenium/generate_config
 
 #=================================
-# Chrome Launch Script Modication
+# Chrome Launch Script Modification
```
testphreak: what changed here?

ddavison:
if this truly fixes these issues (I thought I had fixed the dbus issue..)

then this probably needs to be reverted:

https://github.com/SeleniumHQ/docker-selenium/commit/69707f0ac9ae50d9f810d3a98b5569b3b4e9440b#diff-1454090f007c02ce89531779aca07244R13

I added the "dbus" package

diemol:
Good point @ddavison,
I'll remove the package and test it first, to see if there are any differences. I'll let you know soon.

diemol:
@ddavison I removed the dbus package and executed the tests in the repo and two times this [test](https://gist.github.com/diemol/df6f97b35c3943358a78d8200bbeae35) agains the locally generated containers. So it looks like it works also without the package.

Here are the commands I used to start the containers:
* Generated the local images with:
```
 VERSION="local" make build
```
* docker-compose (for normal and debug, remove the comments where it applies)
```
version: '2'
services:
  selenium-dev:
    image: selenium/hub:3.4.0-local
    ports:
      - 4444:4444/tcp
    environment:
      GRID_BROWSER_TIMEOUT: '180000'
      GRID_MAX_SESSION: '10'
      GRID_TIMEOUT: '180000'
  chrome-dev:
    image: selenium/node-chrome:3.4.0-local
    #image: selenium/node-chrome-debug:3.4.0-local
    #ports:
    #  - 5900:5900/tcp
    environment:
      HUB_PORT_4444_TCP_ADDR: selenium-dev
      HUB_PORT_4444_TCP_PORT: '4444'
      JAVA_OPTS: -Xmx4G
      shm_size: 1024MB
    volumes:
      - /dev/shm:/dev/shm
```
* Standalone images
```
docker run --rm -p 4444:4444 -v /dev/shm:/dev/shm --shm-size=1g selenium/standalone-chrome:3.4.0-local
```
```
docker run --rm -p 4444:4444 -p 5900:5900 -v /dev/shm:/dev/shm --shm-size=1g selenium/standalone-chrome-debug:3.4.0-local
```

Any help testing this is appreciated.

testphreak:
@diemol, not sure I follow. @ddavison in his commit 12 days ago for 3.3.1-cesium, added dbus package to resolve the dbus issues. With the 3.4.0-bismuth, dbus package is no longer needed?

diemol:
@testphreak the dbus package is installed since 3.3.1-cesium, and it is still there for the latest release (3.4.0-bismuth). It attempted to solve the dbus issues, but as we know, they are still there. 

What I did in the last commit was to remove the package and see if it was enough to set `DBUS_SESSION_BUS_ADDRESS=/dev/null`. It seems it is enough, but I would be cool that someone else tests that as well.

I was talking about this with @elgalu, and in my opinion we could try first to just set the value as shown above, since it should solve the issue. 
If after doing that, some people still report the issue (with a reproducible test case), we could install dbus and use it like @elgalu did in his [image](https://github.com/elgalu/docker-selenium), see this [comment](https://github.com/SeleniumHQ/docker-selenium/issues/87#issuecomment-250475864). This would create an extra process in the container though.

testphreak:
@diemol, build locally with your changes and ran a subset of my tests. Looks good to me.

diemol:
Cool @testphreak, thanks for the feedback!

PR #470, opened 2017-04-25
Add NodeFirefoxNightly image

smccarthy:

I get the latest Firefox nightly image, and I also get the latest geckodriver too. I didn't know if it would be best to get the latest GeckoDriver, or have it set in the Dockerfile.

If this is wanted, I could also add the a standalone, standalone debug, and node debug image for nightly as well.


cvakiitho:
Hi @smccarthy , I'd love to use this image ( due to https://github.com/mozilla/geckodriver/issues/858 ), but I'm getting following error when building it: 

```Starting build of index.docker.io/cvakiitho/firefox:latest...
Step 1/20 : FROM selenium/node-base:

invalid reference format
```

Were you able to build this image on docker hub?

diemol:
Hi @smccarthy,

I am sorry that time passed and we didn't give you feedback.

In general, it is a good idea to have the nightly FF as an option. Nevertheless it is not optimal to add one image more to maintain. 

Do you think that it is possible to make this work with the current Dockerfile for FF? Perhaps to have it working in the same way as the one for Chrome, where one can just specify the version as a parameter (such as google-chrome-beta) and then the image gets built. 

Let me know your thoughts,

smccarthy:
@diemol I like that idea. That was actually the feedback I got on Slack when I first opened this PR, but I haven't had the time to update the PR. Maybe this week.

NodeFirefox/Dockerfile.txt
```diff
@@ -4,10 +4,11 @@ USER root
 # Firefox
 #=========
 ARG FIREFOX_VERSION=57.0.2
-RUN apt-get update -qqy \
+RUN FIREFOX_DOWNLOAD_URL=$(if [ $FIREFOX_VERSION = "nightly" ]; then echo "https://download.mozilla.org/?product=firefox-nightly-latest-ssl&os=linux64&lang=en-US"; else echo "https://download-installer.cdn.mozilla.net/pub/firefox/releases/$FIREFOX_VERSION/linux-x86_64/en-US/firefox-$FIREFOX_VERSION.tar.bz2"; fi) \
+  && apt-get update -qqy \
   && apt-get -qqy --no-install-recommends install firefox \
```
smccarthy: I noticed this in there. Is this needed if we get Firefox from Mozilla directly ?

Makefile
```diff
@@ -89,6 +94,14 @@ generate_firefox_debug:
 firefox_debug: generate_firefox_debug firefox
 	cd ./NodeFirefoxDebug && docker build $(BUILD_ARGS) -t $(NAME)/node-firefox-debug:$(VERSION) .
 
+tag_nightly:
+	docker tag $(NAME)/node-firefox:$(VERSION) $(NAME)/node-firefox:nightly
```
smccarthy: Can I delete this tag if I call docker build with -t on line 59?

Issue #458, opened 2017-04-17
Error: rm: cannot remove '/var/cache/apt/archives': Directory not empty


tnguyen14:
Hi, thanks for the fix on #449, it helped me got a bit further when trying to build the docker image locally.

However, I am seeing another error in the build process. Hopefully this one will be easy to fix:

```shell
Setting up openjdk-8-jre-headless:amd64 (8u121-b13-0ubuntu1.16.04.2) ...
update-alternatives: using /usr/lib/jvm/java-8-openjdk-amd64/jre/bin/rmid to provide /usr/bin/rmid (rmid) in auto mode
update-alternatives: using /usr/lib/jvm/java-8-openjdk-amd64/jre/bin/java to provide /usr/bin/java (java) in auto mode
update-alternatives: using /usr/lib/jvm/java-8-openjdk-amd64/jre/bin/keytool to provide /usr/bin/keytool (keytool) in auto mode
update-alternatives: using /usr/lib/jvm/java-8-openjdk-amd64/jre/bin/jjs to provide /usr/bin/jjs (jjs) in auto mode
update-alternatives: using /usr/lib/jvm/java-8-openjdk-amd64/jre/bin/pack200 to provide /usr/bin/pack200 (pack200) in auto mode
update-alternatives: using /usr/lib/jvm/java-8-openjdk-amd64/jre/bin/rmiregistry to provide /usr/bin/rmiregistry (rmiregistry) in auto mode
update-alternatives: using /usr/lib/jvm/java-8-openjdk-amd64/jre/bin/unpack200 to provide /usr/bin/unpack200 (unpack200) in auto mode
update-alternatives: using /usr/lib/jvm/java-8-openjdk-amd64/jre/bin/orbd to provide /usr/bin/orbd (orbd) in auto mode
update-alternatives: using /usr/lib/jvm/java-8-openjdk-amd64/jre/bin/servertool to provide /usr/bin/servertool (servertool) in auto mode
update-alternatives: using /usr/lib/jvm/java-8-openjdk-amd64/jre/bin/tnameserv to provide /usr/bin/tnameserv (tnameserv) in auto mode
update-alternatives: using /usr/lib/jvm/java-8-openjdk-amd64/jre/lib/jexec to provide /usr/bin/jexec (jexec) in auto mode
Processing triggers for libc-bin (2.23-0ubuntu7) ...
Processing triggers for systemd (229-4ubuntu16) ...
[...]

PR #480, opened 2017-05-08
Use `apt-get clean`

tnguyen14:
Instead of `rm -rf /var/cache/apt/archives`, which causes an error on
RHEL 7.3 (see SeleniumHQ/docker-selenium#458), do `apt-get clean`
instead

### DIFF
--- a/Base/Dockerfile
+++ b/Base/Dockerfile
@@ -21,7 +21,8 @@ RUN apt-get update -qqy \
     sudo \
     unzip \
     wget \
-  && rm -rf /var/lib/apt/lists/* /var/cache/apt/* \
+  && rm -rf /var/lib/apt/lists/* \
+  && apt-get clean \
   && sed -i 's/securerandom\.source=file:\/dev\/random/securerandom\.source=file:\/dev\/urandom/' ./usr/lib/jvm/java-8-openjdk-amd64/jre/lib/security/java.security
 
 #===================
--- a/NodeBase/Dockerfile.txt
+++ b/NodeBase/Dockerfile.txt
@@ -8,7 +8,8 @@ RUN apt-get update -qqy \
   && apt-get -qqy install \
     locales \
     xvfb \
-  && rm -rf /var/lib/apt/lists/* /var/cache/apt/*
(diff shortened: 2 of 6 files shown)


diemol:
Hi @tnguyen14,

I have time now to check this, could you please update your fork and submit the changes again?

tnguyen14:
Hi yes, I will update and re-submit. Been a bit overly busy at work these days, will take a stab at this as soon as I can! Thanks for checking in.

alistair23:
Looking at https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#run it seems you don't need to either rm the packages or use apt-get clean as the base Ubuntu image will do this for you.

ddavison:
not sure we need apt-get clean. 

the `rm -fr` command is plenty.

In the link you shared, @alistair23 - even says:

> Below is a well-formed RUN instruction that demonstrates all the apt-get recommendations.

```Dockerfile
RUN apt-get update && apt-get install -y \
    aufs-tools \
    automake \
    build-essential \
    curl \
    dpkg-sig \
    libcap-dev \
    libsqlite3-dev \
    mercurial \
    reprepro \
    ruby1.9.1 \
    ruby1.9.1-dev \
    s3cmd=1.1.* \
 && rm -rf /var/lib/apt/lists/*
```

They recommend rm -fr'ing the apt lists directory, which we are already doing.

diemol:
Well, this PR was open because the build didn't work in a given environment, but it has been abandoned. I think I'll close it and perhaps it can be submitted again in the future, if it makes sense.

tnguyen14:
So is the right fix to remove `rm /var/cache/apt` instead of using `apt-get clean`?

What is the reason to not use `apt-get clean`?

alistair23:
We don't need to use either. There is a note on that page that says

> Note: The official Debian and Ubuntu images automatically run apt-get clean, so explicit invocation is not required.
See here: https://github.com/moby/moby/blob/03e2923e42446dbb830c654d0eec323a0b4ef02a/contrib/mkimage/debootstrap#L82-L105

and as the images are based on Ubuntu the rm -rf and apt-get clean are unnecessary.

PR #494, opened 2017-06-02
get chrome version from chrome -version

jonaseicher:
this retrieves the version without the need for sudo, which causes problems when you deploy in a container environment with tighter security guidelines like openshift.

Feel free to refine/improve it or use a different approach. The main thing here is not to use sudo.


ddavison:
definitely like this solution over `sudo`.   had problems in the past when using openshift because of the security.

PR #185, opened 2016-03-22
move the creation of the config.json file into a RUN directive

ddavison:
You would *think* that this is a no-op, but it's not.  This was actually creating an issue when running Docker containers in succession with Kubernetes and OpenShift.

This change won't affect other users using this without Kubernetes and OpenShift.  Due to the nature that some Docker containers won't be safe - OpenShift enforces a policy of an [arbitrary user ID](https://docs.openshift.com/enterprise/3.0/creating_images/guidelines.html) that has sudo'ers rather than running as root.

PR #502, opened 2017-06-19
Moving config generation in docker files

Remi-p:
Hi,
Following the issue [Permission denied issue when starting node-chrome pod #3](https://github.com/ddavison/selenium-openshift-templates/issues/3#issuecomment-306546091) from ddavison/selenium-openshift-templates, I propose to make the same modification type as in [move the creation of the config.json file into a RUN directive #185](https://github.com/SeleniumHQ/docker-selenium/pull/185/files).

Typically, moving this line from `NodeBase/entry_point.sh` :
```
/opt/selenium/generate_config > /opt/selenium/config.json
```
To the end of the dockerfiles (with `RUN` directive).

`make generate_all`, `make all`, & `make test` have been run. Also I tried the built node-chrome image with OpenShift, it works (see [this temporary Docker Hub image](https://hub.docker.com/r/ppersonne/node-chrome-verif/)).

### DIFF
--- a/NodeBase/entry_point.sh
+++ b/NodeBase/entry_point.sh
@@ -1,7 +1,6 @@
 #!/bin/bash
 
 source /opt/bin/functions.sh
-/opt/bin/generate_config > /opt/selenium/config.json
 
 export GEOMETRY="$SCREEN_WIDTH""x""$SCREEN_HEIGHT""x""$SCREEN_DEPTH"
 
--- a/NodeChrome/Dockerfile
+++ b/NodeChrome/Dockerfile
@@ -46,3 +46,6 @@ COPY generate_config /opt/bin/generate_config
 # Chrome Launch Script Modification
 #=================================
 COPY chrome_launcher.sh /opt/google/chrome/google-chrome
+
+# Generating config inside the image rather than with entry_point.sh
+RUN /opt/bin/generate_config > /opt/selenium/config.json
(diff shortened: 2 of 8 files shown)


ddavison:
yep!  i'm good with this.  please do me favor though, and take out your `Dockerfile` changes?  please only change the `Dockerfile.txt`.  In the header of `Dockerfile`, it kindly specifies not to edit it.  This is because this file is generated.

I'll merge this in as soon as that is fixed!  thanks @Remi-p !  nice work.

Remi-p:
My bad I thought the warning was for local modifications...
Should I rebase the two commits together or is it OK ?

ddavison:
don't worry about squashing.  thanks @Remi-p !

Remi-p:
Hey ! Sorry to bother you, but should I do another modification ? I say that because of the awaiting-external-changes label.
Thanks for your reply !

ddavison:
oops, thanks for reminding me!

Remi-p:
No problem!

Remi-p:
It was for correcting a permission problem (see [Permission denied issue when starting node-chrome pod](https://github.com/ddavison/selenium-openshift-templates/issues/3)). But you're right this way it breaks the use of environment variables.
A solution might be to generate the config file at build time, like it is done here, and regenerate it at run time ? As it is done with the Hub. This way, people having permissions problems will have a default configuration, and the ones wanting to customize the variables will have the opportunity.

PR #529, opened 2017-08-03
Config generation at runtime + buildtime

Remi-p:
Hi,
As @junhui pointed out, my previous pull request (#502) broke the ability to pass environment variables for config generation.

I believe a fix could be to generate config at build time (it creates a default config file), and at runtime. As it is actually done with the Hub : [Hub/Dockerfile#L39](https://github.com/SeleniumHQ/docker-selenium/blob/1fbe302c398d45dd16a1faa6929ebc0811741f89/Hub/Dockerfile#L39) and [Hub/entry_point.sh#L6](https://github.com/SeleniumHQ/docker-selenium/blob/0f12883b3deac2a1bc63552a98f51704a07e288c/Hub/entry_point.sh#L6). See [issue 502 last comment](https://github.com/SeleniumHQ/docker-selenium/pull/502#issuecomment-318816235).

### DIFF
--- a/NodeBase/entry_point.sh
+++ b/NodeBase/entry_point.sh
@@ -1,6 +1,7 @@
 #!/bin/bash
 
 source /opt/bin/functions.sh
+/opt/bin/generate_config > /opt/selenium/config.json
 
 export GEOMETRY="$SCREEN_WIDTH""x""$SCREEN_HEIGHT""x""$SCREEN_DEPTH"
 
--- a/NodeFirefoxDebug/entry_point.sh
+++ b/NodeFirefoxDebug/entry_point.sh
@@ -3,6 +3,7 @@
 # IMPORTANT: Change this file only in directory NodeDebug!
 
 source /opt/bin/functions.sh
+/opt/bin/generate_config > /opt/selenium/config.json
 
 export GEOMETRY="$SCREEN_WIDTH""x""$SCREEN_HEIGHT""x""$SCREEN_DEPTH"
 
--- a/NodeChrome/Dockerfile.txt
+++ b/NodeChrome/Dockerfile.txt
@@ -40,5 +40,5 @@ COPY generate_config /opt/bin/generate_config
 #=================================
 COPY chrome_launcher.sh /opt/google/chrome/google-chrome
 
-# Generating config inside the image rather than with entry_point.sh
+# Generating a default config during build time
 RUN /opt/bin/generate_config > /opt/selenium/config.json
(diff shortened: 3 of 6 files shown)


NodeBase/entry_point.sh
```diff
@@ -1,6 +1,7 @@
 #!/bin/bash
 
 source /opt/bin/functions.sh
+/opt/bin/generate_config > /opt/selenium/config.json
```
ddavison: `NodeBase` doesn't have a `generate_config` this would fail, no?

NodeFirefoxDebug/entry_point.sh
```diff
@@ -3,6 +3,7 @@
 # IMPORTANT: Change this file only in directory NodeDebug!
 
 source /opt/bin/functions.sh
+/opt/bin/generate_config > /opt/selenium/config.json
```
ddavison: Would these lines replace the `RUN /opt/bin/generate_config > /opt/selenium/config.json` lines? or would they be supplemental

NodeBase/entry_point.sh
Remi-p: Before the commit 0dd193a (*Moving config generation in docker files*) the line was here.

I believe that it's because NodeBase is not runned directly, it's used in other images that have generate_config.

---

For instance for Chrome.
In NodeBase [Dockerfile.txt](https://github.com/SeleniumHQ/docker-selenium/blob/master/NodeBase/Dockerfile.txt#L56), the entry point is set :
```
CMD ["/opt/bin/entry_point.sh"]
```
It calls `entry_point.sh` which calls for `generate_config`.

In NodeChrome [Dockerfile.txt](https://github.com/SeleniumHQ/docker-selenium/blob/master/NodeChrome/Dockerfile.txt#L36), the `generate_config` file is copyied inside the image, before the entrypoint is called :
```
COPY generate_config /opt/bin/generate_config
```

NodeFirefoxDebug/entry_point.sh
Remi-p: I think they should both exist.
The RUN directive would create a default configuration file, and the generate_config from the `entry_point.sh` will allow to customize environment variables.

This is actually done for the hub : [Hub/Dockerfile#L39](https://github.com/SeleniumHQ/docker-selenium/blob/master/Hub/Dockerfile#L39) & [Hub/entry_point.sh#L6](https://github.com/SeleniumHQ/docker-selenium/blob/master/Hub/entry_point.sh#L6)

PR #545, opened 2017-08-31
Use latest chromedriver version by default

glib-briia:
Use latest chromedriver version, since latest stable Chrome version is used by default.
CHROME_DRIVER_VERSION still can be overridden when building the image.

### DIFF
--- a/NodeChrome/Dockerfile.txt
+++ b/NodeChrome/Dockerfile.txt
@@ -21,17 +21,22 @@ RUN wget -q -O - https://dl-ssl.google.com/linux/linux_signing_key.pub | apt-key
 
 USER seluser
 
-#==================
+#============================================
 # Chrome webdriver
-#==================
-ARG CHROME_DRIVER_VERSION=2.31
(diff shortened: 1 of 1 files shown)


ddavison:
i like the idea, and a good implementation!

I think i'll wait to merge this with advice from the community / team though, since there are potential drawbacks to something like this.

As with any dependency, you shouldn't always fetch latest because of compatibility issues.  I'd hate to run into an issue where people are downloading 3.5.1 (later in the future) and they can't run it without building their own containers because the version is fetching latest.

There's always potential drawbacks to doing something like this.

On the other hand, we already do this for things like google-chrome.  we use an `apt-get -qqy install \
    ${CHROME_VERSION:-google-chrome-stable}` which basically does the same thing.

glib-briia:
Hi @ddavison ,

Thank you for the feedback. Yes, as you have said the 'latest' approach is already used for Chrome and since its tightly coupled with chromedriver its either use 'latest' for both or manually simultaneously update versions for both imho. 

Surely better to discuss wit community first.

Thanks

hazmeister:
Looks like a sensible default, it's still possible to override with a specific version if you need it.

ddavison:
thanks! 👍

PR #563, opened 2017-09-11
Fix for issue with not being able to pass environment variables into node containers at runtime

testphreak:
When changes were introduced to generate default config during build
time with #502 and then with #529, it broke the ability to pass in
environment variables to the node containers at runtime. This fix aims
to provide the ability to do both i.e. continue to pass environment
variables at runtime and generate a default config at build time for
the node.

### DIFF
--- a/NodeBase/Dockerfile.txt
+++ b/NodeBase/Dockerfile.txt
@@ -30,26 +30,6 @@ ENV SCREEN_HEIGHT 1020
 ENV SCREEN_DEPTH 24
 ENV DISPLAY :99.0
 
-#========================
-# Selenium Configuration
-#========================
-# As integer, maps to "maxInstances"
-ENV NODE_MAX_INSTANCES 1
-# As integer, maps to "maxSession"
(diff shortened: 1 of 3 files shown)


testphreak:
Hi @Remi-p, @ddavison, @diemol - please review this change and merge if it looks good. 

Unsure why CI has been failing since build 315. Was able to build and test my change locally.

diemol:
Hi @testphreak,

Seems something is wrong with the tests, were you able to run the build script locally?

testphreak:
Hi @diemol, yes, I was able to run the build script locally. There appears to be something else wrong that maybe specific to the CI environment.

testphreak:
Didn't realize that there was already a fix to the environment variables issue and was awaiting a new release of the images. Closing this PR.

glib-briia:
Fix for the build is here #561

testphreak:
Hi @diemol and @ddavison could we get #561 merged so we can fix the build and release new images to fix passing environment variables at runtime?

Issue #599, opened 2017-10-31
Fails to build standalone-chrome-debug (missing node-chrome-debug dependency)


eitany:
## Meta -
Image(s):  
standalone-chrome-debug
Docker-Selenium Image Version(s):  
3.6.0-copper
Docker Version:  
1.8.2
OS: 
OSX Yosemite

## Actual Behavior -
Docker build fails, missing dependency node-chrome-debug

diemol:
Hi @eitany,

Good catch, I was able to reproduce it. I made some comments to your PR #600, please solve them and we can merge.

Thanks!

diemol:
#600 was merged, so closing this issue.
Thanks @eitany

PR #600, opened 2017-10-31
Add missing chrome_debug dependency

eitany:

Fix for: https://github.com/SeleniumHQ/docker-selenium/issues/599

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -75,7 +75,7 @@ standalone_chrome: generate_standalone_chrome chrome
 generate_standalone_chrome_debug:
 	cd ./StandaloneDebug && ./generate.sh StandaloneChromeDebug node-chrome-debug Chrome $(VERSION) $(NAMESPACE) $(AUTHORS)
 
-standalone_chrome_debug: generate_standalone_chrome_debug standalone_chrome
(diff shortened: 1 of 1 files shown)


diemol:
Hi @eitany,

Thanks for this PR, good catch!

Makefile
```diff
@@ -75,7 +75,7 @@ standalone_chrome: generate_standalone_chrome chrome
 generate_standalone_chrome_debug:
 	cd ./StandaloneDebug && ./generate.sh StandaloneChromeDebug node-chrome-debug Chrome $(VERSION) $(NAMESPACE) $(AUTHORS)
 
-standalone_chrome_debug: generate_standalone_chrome_debug standalone_chrome
+standalone_chrome_debug: chrome_debug generate_standalone_chrome_debug standalone_chrome
```
diemol: I think you just need to replace `standalone_chrome` with `chrome_debug`, instead of adding it. 
Could you please change that?

I also noted that the `standalone_firefox_debug` has the same issue, if you want to, make that change as well.

diemol: I'll merge it like that and then make the additional change. Thanks again!

PR #614, opened 2017-11-22
pulseaudio package installation for Firefox containers

pabloFuente:

Hi. First of all, sorry for the previous PR. You're right: the file clearly says not to modify it.

I have already shifted the change to Dockerfile.txt on NodeFirefox, so all StandaloneFirefox, StandaloneFirefoxDebug, NodeFirefox and NodeFirefoxDebug should now include the change.

Let me dig a little deeper into this PR:

  - At OpenVidu ([https://github.com/OpenVidu/openvidu](https://github.com/OpenVidu/openvidu)) we work with media transmission over the Internet thanks to [WebRTC standard](https://webrtc.org/). Due to the changing nature of the framework, we need a good test coverage able to automatically test the video transmission in an e2e scenario for both Chrome and Firefox. We use Selenium Docker containers to do so.
  - We have checked that there's no problem at all with Selenium's Chrome images: media transmission is succesful. But in Firefox we found out that tests were failing, accompanied by the always uncomfortable situation of no having any error or warning messages at all. After doing some reasearch, we came to the conclusion that the problem was the standard method ['getUserMedia'](https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices/getUserMedia) only in Firefox, along with the lack of **pulseaudio** package in the container (this is the bugzilla official entry: [https://bugzilla.mozilla.org/show_bug.cgi?id=1247056](https://bugzilla.mozilla.org/show_bug.cgi?id=1247056)).
  - To be sure about this fix, we have already checked that the same test running on Selenium Firefox container fails if the container is not running pulseaudio but passes if it is doing so.

Output in the first container when checking if pulseaudio is running:

```
seluser@ff937bb8dcf7:/$ pulseaudio -D
bash: pulseaudio: command not found
seluser@ff937bb8dcf7:/$
```

RESULT: Test fails

Output in the second container:

```
seluser@0d6ec579fcfa:/$ pulseaudio -D
seluser@0d6ec579fcfa:/$ 
```

(pulseaudio returns exit 0 for this command when it is properly running)
RESULT: Test passes

  - When would this package be used? Every time any end2end test launched against Selenium's Dockerized Firefox needs to establish a WebRTC connection that includes audio. This [StackOverflow entry](https://stackoverflow.com/questions/44775411/getusermedia-in-firefox-in-docker-not-working-when-using-audio) explains in detail this bug and how to reproduce it using an offical WebRTC Demo. We have already tested our custom scenario and the scenario from this StackOverflow's entry. With this change built in the Firefox container it works prefectly ok.

Thanks for your time.


NodeFirefox/Dockerfile.txt
```diff
@@ -38,3 +38,7 @@ RUN sudo echo ""
 
 # Generating a default config during build time
 RUN /opt/bin/generate_config > /opt/selenium/config.json
+
```
diemol: I wonder if it is better to install the package when the user is root and avoid using sudo... If you install it when the user is root, are the results the same @pabloFuente? 
If so, I think it would be better to install it with root.

diemol:
Hi @pabloFuente,

Thanks for submitting the PR again, also thank you for the clear explanation for it. 

Did you check how much does the image grow by installing the package? I guess not so much, would be nice to know that.

From my point of view, it looks good. I don't see why we could not merge it.

@ddavison, what do you think?

pabloFuente:
Hello again. I am glad to answer your questions:
- The size of the image grows from 740 MB to 790 MB.
- Your requested change is pretty convenient. I have already commited it: now package is installed on user *root* without sudo instead of in *seluser*. Everything's still working fine, as before.

Thank you.

diemol:
Thanks for the changes!

From my point of view it looks good. The extra 50MB is not ideal but I don't think there is a big issue about it. The Chrome container is already 848MB for example. We should look into that later.

I'll just wait for @ddavison's opinion, after that, it should be ok to merge.

ddavison:
i guess i'm just confused of the practicality of this change.  i appreciate the deep explanation, @pabloFuente but i guess i'm still just confused of the practicality.  If we installed this in the containers...

- how often would this be used?
- by whom?  (other than yourselves)
- In a selenium test, do many people test this sort of audio transmission?  seems like Selenium wouldn't be the best software for something like that.

I noticed you had only changed the NodeFirefox image.  Does this already "passively" exist in the NodeChrome images?  If this is the case, then I have no problem with it whatsoever.  If doing an "apt-get google-chrome-stable..." automatically installs pulseaudio, then we're green for this PR.

bonigarcia:
Hi @ddavison. If I may, I would like to put in my two cents here.

The whole point of this PR is to made available Firefox for testing WebRTC applications using dockerized browsers. I trace the problem in the past, as explained in the [stackoverflow question](https://stackoverflow.com/questions/44775411/getusermedia-in-firefox-in-docker-not-working-when-using-audio) previously linked by @pabloFuente. The problem started in Firefox 52. It seems the internal implantation of the WebRTC stack requires pulseaudio to work as of this version.

Regarding your first and second questions, it is required by people testing WebRTC applications and using the latest versions of Firefox by means of Docker. Regarding your third question, I disagree with your comment. Selenium is perfectly capable of that. In fact, it is a must to make it in an automated fashion, since the WebRTC stack is mainly implemented in modern real browsers such as Chrome or Firefox, and therefore, to test these services properly, we need to drive programmatically browsers (i.e. using Selenium). Here, in my opinion it is very interesting to use dockerized browsers, for example to carry out compatibility tests (i.e. different version or browser type) or performance test (i.e. use a lot of browsers in a test), but at this moment it is not possible using the official Selenium Docker images, which is a pity for the open source community.

Regarding your last comment, no, Chrome does not require pulseaudio. I suppose this is because the internal implementation of the WebRTC is different in Chrome. But IMHO this should not affecting the final decision about merging this PR.

ddavison:
your comments are very welcome, @bonigarcia!  thanks 👍 

thanks for also greater clarification.  perhaps it's my ignorance on WebRTC.  

>  Regarding your first and second questions, it is required by people testing WebRTC applications and using the latest versions of Firefox by means of Docker.

Ok, so this is a Firefox specific thing.  perfect.

> Regarding your third question, I disagree with your comment. Selenium is perfectly capable of that

Well, i didn't say it wasn't capable, i just said it probably wasn't the "best" option.  again, this may stem from my lack of knowledge of WebRTC.


As a summation, i've done a little more research, and these testers that need this - are testing sites that use WebRTC.  This includes, perhaps sites that use the computer peripherals (camera, mic, etc).

**I think it's perfectly reasonable to include this.**

@pabloFuente , looks like your branch is out of date.  I still don't trust the "Update branch" button in 
GitHub so I'm going to ask you to update your branch with the latest changes and ensure there are no conflicts.

bonigarcia:
@ddavison Cool, thanks to you!

pabloFuente:
There it goes. Updated to master.

Issue #618, opened 2017-11-24
selenium/standalone-chrome:3.7.1-beryllium inconsistent start


phuongnd08:
We use selenium/standalone-chrome:3.7.1-beryllium on Circle to run some test. The container starts most of the time but randomly failed the other time. Please advice on how to debug this issue (or advice about a version that consistently starts. We don't mind an outdated browser)

## Meta -
Image(s):  
<img width="605" alt="screen shot 2017-11-24 at 7 29 36 am" src="https://user-images.githubusercontent.com/184037/33191201-531cdfbc-d0e9-11e7-8bf4-48aa88f649c7.png">

Docker-Selenium Image Version(s):  
selenium/standalone-chrome:3.7.1-beryllium
Docker Version:  
Not sure, default Circle CI provided version
OS: 
Linux

## Expected Behavior -
The container launch consistently
## Actual Behavior -
The container fail randomly (1 out of 10 times)

diemol:
Hi @phuongnd08,

Can you please share the command you used to start the containers?

phuongnd08:
We are using default command (the entry)

diemol:
Oh, I was looking for something more specific, to see the `docker run...` command being used.

Also, taking the [comment](https://github.com/SeleniumHQ/docker-selenium/pull/619#issuecomment-346807718) from #619
>I already did this. Though I would like to think having such change will make it easier for people to report and diagnose the issue (which can come in a variety of forms)

I understand that you created an image based on the Standalone and modified it to look like:
```sh
xvfb-run -e /dev/stdout -n $SERVERNUM --server-args="-screen 0 $GEOMETRY -ac +extension RANDR" \
    java ${JAVA_OPTS} -jar /opt/selenium/selenium-server-standalone.jar \		    java ${JAVA_OPTS} -jar /opt/selenium/selenium-server-standalone.jar \
    ${SE_OPTS} &		    ${SE_OPTS} &
```

Did you get enough output that helps to find the cause why the `Xvfb failed to start` happens?

phuongnd08:
After adding such params we have yet to observe the issue again. Will report if it happen.

phuongnd08:
Okay, with that command all we get is this: 
```
xvfb-run: error: Xvfb failed to start

Exited with code 1
```
Do you have recommend on a flag that would dump more the error?
[...]

PR #619, opened 2017-11-24
Add -e /dev/stdout to xvfb-run

phuongnd08:
So when Xvfb failed to start we have a remote ideas of what went wrong.

### DIFF
--- a/Standalone/entry_point.sh
+++ b/Standalone/entry_point.sh
@@ -19,7 +19,7 @@ SERVERNUM=$(get_server_num)
 
 rm -f /tmp/.X*lock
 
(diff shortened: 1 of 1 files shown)


diemol:
Hi @phuongnd08,

I suggest that you create an image with these changes first and see what output this brings in the environment you are using it. Then perhaps the root cause will be seen and an improved PR can be sent.

Otherwise, we would be the bottleneck in your pipeline, since we would need to merge, release and then this would not really guarantee that it works. 

I'll close this PR and we can follow up on #618 based on your findings.

phuongnd08:
I already did this. Though I would like to think having such change will make it easier for people to report and diagnose the issue (which can come in a variety of forms)

PR #625, opened 2017-11-30
Install Python2, Python2 PIP and Selenium for Python in the base image

alistair23:

### DIFF
--- a/Base/Dockerfile
+++ b/Base/Dockerfile
@@ -17,6 +17,7 @@ ENV DEBIAN_FRONTEND=noninteractive \
 # Includes minimal runtime used for executing non GUI Java programs
 #========================
 RUN apt-get -qqy update \
+  && apt-get upgrade -y \
   && apt-get -qqy --no-install-recommends install \
     bzip2 \
     ca-certificates \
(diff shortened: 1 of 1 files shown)


diemol:
Hi @alistair23,

Thanks for sending us a PR. Could you please tell us what is the need to install Python in the images?

alistair23:
I use Selenium in Docker containers by running Python scripts. I thought it might be beneficial to install python with selenium support in the base image.

diemol:
Well, yes, it would benefit users who have tests in python. But this would leave out users who use ruby, java, etc...

Bottomline, the images try to be a generic as possible, and this would go against that concept. That's why I think we cannot accept this PR, sorry.

alistair23:
That's fair. Is there any plan on having larger more complete images or maybe specific development images in the future?

diemol:
Not really, at least not officially. You could use this one https://github.com/elgalu/docker-selenium, which already has Python installed (but it is a heavier image).

I'll close the PR :)

alistair23:
No worries.
Is the first patch (just the upgrade) ok to be merged?

diemol:
Hm, maybe, I am not sure about the benefit. But you could send the PR and then we discuss it also with @ddavison

PR #626, opened 2017-11-30
Upgrade packages in base image

alistair23:
There are numerous upgrades available to the Ubuntu base packages. These range from security updates to bug fixes. It makes sense to upgrade these packages to patch any security vulnerabilities and fix any potential bugs.

The Ubuntu LTS packages are usually very conservative, so there is limited (although possibly) some risk of breakages. This risk already exists for all installed packages though, as they will take the newest version.

### DIFF
--- a/Base/Dockerfile
+++ b/Base/Dockerfile
@@ -17,6 +17,7 @@ ENV DEBIAN_FRONTEND=noninteractive \
 # Includes minimal runtime used for executing non GUI Java programs
 #========================
 RUN apt-get -qqy update \
+  && apt-get upgrade -y \
   && apt-get -qqy --no-install-recommends install \
     bzip2 \
     ca-certificates \


diemol:
Hi @alistair23,

The images will grow 8MB, which is not really an issue. 

I researched a bit, and I don't know if it makes sense to include this. From the Dockerfile RUN [best practices](https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#run):  
>You should avoid `RUN apt-get upgrade` or `dist-upgrade`, as many of the “essential” packages from the parent images won’t upgrade inside an unprivileged container. If a package contained in the parent image is out-of-date, you should contact its maintainers. If you know there’s a particular package, `foo`, that needs to be updated, use `apt-get install -y foo` to update automatically.

So this suggests that we should only update packages when we know that it is needed. In addition, the base image is `ubuntu:16.04` and they tend to release once or twice a month with improvements and security updates, and since we release several times per month, we are getting those updates implicitly. 

Initially I thought this was a good idea, but after reading docs and docker forum's entries, I understand that it does not make sense.

If you are still eager to contribute, please help us by checking the open issues and tacking one of them. We could point you how to get started in any case. Another good point is to update the documentation in the README.

I'll close the PR for the reasons mentioned above.

alistair23:
Thanks for pointing that out. I didn't realize that updating was against the best practices. I've definitely seem it done all over the place. Thanks for letting me know.

I'll see if there are any issues I think I can help with.

PR #639, opened 2017-12-15
Bringing back support for -remoteHost

diemol:
--remoteHost was added back in the server in 3.7.1
This is useful when the containers are deployed in different VMs and under special network conditions.


ddavison:
ref: https://github.com/seleniumhq/selenium/commit/827bbd8177651a9316a574c9622f7d1226221173

Issue #367, opened 2017-01-17
Chinese garbled in node-firefox and node-chrome when screen shot


zhengchunxi:
## Meta -
Image(s):  
selenium/node-firefox and selenium/node-chrome
Docker Version:  
3.0.1
OS: 
ubuntu 14.04 x86_64

## Expected Behavior -
Chinese show normal
## Actual Behavior -
Chinese garbled

I use my Physical device as a node,Chinese show normal
but use docker image, Chinese garbled

elgalu:
probably related to #339

individual-it:
same problem with Devanagari script
![image](https://user-images.githubusercontent.com/2425577/34718320-4bae844e-f55e-11e7-9016-b82328f64cb2.png)

diemol:
This should be solved in the latest release
https://github.com/SeleniumHQ/docker-selenium/releases/tag/3.11.0-bismuth

Somehow fonts were only being installed in the debug images, now they are available for both types of images.

Please reopen or comment if the issue persists.

PR #676, opened 2018-02-23
Add Chinese fonts for #339 #367

omz-one:
Add Chinese fonts ttf-wqy-microhei and ttf-wqy-zenhei

- [X ] By placing an `X` in the preceding checkbox, I verify that I have signed the [Contributor License Agreement](https://github.com/SeleniumHQ/docker-selenium/blob/master/CONTRIBUTING.md#contributing-code-to-selenium)

### DIFF
--- a/Base/Dockerfile
+++ b/Base/Dockerfile
@@ -25,6 +25,8 @@ RUN apt-get -qqy update \
     sudo \
     unzip \
     wget \
+    ttf-wqy-microhei \
+    ttf-wqy-zenhei \
   && rm -rf /var/lib/apt/lists/* /var/cache/apt/* \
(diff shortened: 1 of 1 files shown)


omz-one:
img

diemol:
He @omz-one,

Thanks for this PR! Did you check how much bigger the image gets?

diemol:
Hi @omz-one,

Could you check my comment? In addition, why is the change made on the Base dockerfile? Shouldn't it be done on the NodeBase dockerfile?

diemol:
OK, so the images grow around ~25MB, which is ok I think, considering that a bigger user base will be able to render websites that use this character set.

We'll merge it but then move the two fonts to the NodeBase dockerfile.

diemol:
Before:
before
After:
after

PR #679, opened 2018-02-25
Build Firefox developer edition

### LINKED ISSUE #536: Build Firefox Developer Edition
First off, thanks for all of your work on these images. They have helped us immensely for our testing in CI.

I am trying to test web extension through Protractor and Selenium using these images. Firefox won't run the addon, because it is a folder and unsigned. In [order to run an unsigned extension in Firefox, you can't use the main Firefox release, but can use the developer edition with a custom setting](https://support.mozilla.org/en-US/kb/add-on-signing-in-firefox#w_what-are-my-options-if-i-want-to-use-an-unsigned-add-on-advanced-users). Would it be possible to have images with the Firefox Developer Edition, so that we can test extensions on them?

I am happy to submit a pull request with this functionality, if it's something you are interested in.

l0b0:
Resolves #536.

### DIFF
--- a/NodeFirefox/Dockerfile.txt
+++ b/NodeFirefox/Dockerfile.txt
@@ -4,7 +4,7 @@ USER root
 # Firefox
 #=========
 ARG FIREFOX_VERSION=58.0.2
-RUN FIREFOX_DOWNLOAD_URL=$(if [ $FIREFOX_VERSION = "nightly" ]; then echo "https://download.mozilla.org/?product=firefox-nightly-latest-ssl&os=linux64&lang=en-US"; else echo "https://download-installer.cdn.mozilla.net/pub/firefox/releases/$FIREFOX_VERSION/linux-x86_64/en-US/firefox-$FIREFOX_VERSION.tar.bz2"; fi) \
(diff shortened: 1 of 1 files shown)


diemol:
Hi @l0b0,

Thanks for this PR!

Nevertheless, we would not push a new image with Firefox Developer Edition to Docker Hub. The current images try to be as generic as possible, and this is not a common use case. The idea is that it should work in the same way as the Nightly or Chrome Beta, in the sense that if anyone needs that image, s/he should build it on her/his own.

Having said that, this is the [part](https://github.com/l0b0/docker-selenium/blob/ebff445638d3b4782fb08549fcce50ddbc701554/NodeFirefox/Dockerfile.txt#L7) that we would accept. Could you please modify the PR so it only includes that? Thanks!

l0b0:
Looks like the build has to be re-run; it timed out for some reason.

ddavison:
LGTM 👍

denishowew:
An image for Firefox developer edition would let us test our browser extension without having to go through Mozilla's slow and flaky signing process every time.  Dev edition is the only way to load unsigned extensions.

diemol:
As mentioned above, we want to offer generic images, and aside from that, we do not have the capacity to maintain an image with a browser that has not been released yet. If issues are raised to its behavior, only the time spent to triage them is already an overhead for us. That is why we offer instructions for the community on how to do it, and with that unblock themselves, anyone can set a CI env and build that image on a daily basis if wished.

Issue #537, opened 2017-08-11
Chrome does not respect http proxy environment variables


tnguyen14:
## Meta -
Image(s):  
node-chrome, node-chrome-debug
Docker-Selenium Image Version(s):  
3.4
OS: 
RHEL7.3

Chrome does not respect `http_proxy`, `https_proxy` and `no_proxy` environment variables that are set on the docker image. If these properties are passed to webdriver settings, chrome does respect them however.

I first asked about this in #478.

diemol:
Hi @tnguyen14,

Can you please share the docker command you are using to start the images?
How are you able to see that those env vars are not respected?

AndreSteenbergen:
Latest version actually does. I have spun up 2 azure machines, one with squid proxy, one without. Following that I created 2 nodes with different applicationName (NoProxy and WithProxy). After which I created a test console app, To first browse to https://whatismyipaddress.com/ and create a screenshot. Both screenshots showed a different ip address.

these are working settings (copy pasted from rancher)

![image](https://user-images.githubusercontent.com/7676790/32413864-8f44d608-c21a-11e7-9685-c25207a9fd89.png)

AndreSteenbergen:
Without proxy:
![image](https://user-images.githubusercontent.com/7676790/32413990-13f83794-c21d-11e7-80b8-ab92510f8a4b.png)


With Proxy
![image](https://user-images.githubusercontent.com/7676790/32413983-edef8c3c-c21c-11e7-8499-e9bd4482fd4a.png)

AndreSteenbergen:
I actually don't know which are the correct env variables, with or without CAPS ...

AndreSteenbergen:
Test script: (c#)
I created a custom image with ublock standard enabled, but that shouldn't be too big of a difference (just extra files under the /tmp/ folder.) 

```
Console.WriteLine("Hello World!");

            ChromeOptions options = new ChromeOptions();
            options.AddArguments("disable-infobars");
            options.AddArguments("load-extension=/tmp/ublock/");

            var dc = options.ToCapabilities() as DesiredCapabilities;
            dc.SetCapability("applicationName", "WithProxy");            

            var prop = dc.GetType().GetField("isSpecCompliant", System.Reflection.BindingFlags.NonPublic | System.Reflection.BindingFlags.Instance);
            prop.SetValue(dc, false);

[...]

Issue #78, opened 2015-07-17
Chrome and proxy environment variables


cristian-smocot:
Chrome doesn't respect ``http_proxy`` and`` https_proxy`` environment variables.

I'm using the variables like this
```
https_proxy=http://user:password@proxycorp.com:8080
http_proxy=http://user:password@proxycorp.com:8080
```

I've also tried:
```
https_proxy=user:password@http://proxycorp.com:8080
http_proxy=user:password@http://proxycorp.com:8080
```
to no avail.

I've also used the uppercase variants like HTTP_PROXY and HTTPS_PROXY, with no success.

Chrome doesn't respect these environment variables.
In the debug images, I used bash to see if my environment variable exist, and they do. When I start the container in `ENV` I put the proxies environment variables.

cristian-smocot:
In case anyone is interested for a fix or workaround this, here is what I did:
 - I Installed these extensions on Chrome
  * [__Proxy Auto Auth__](https://chrome.google.com/webstore/detail/proxy-auto-auth/ggmdpepbjljkkkdaklfihhngmmgmpggp?hl=en)
  * [Give Me CRX](https://chrome.google.com/webstore/detail/give-me-crx/acpimoebmfjpfnbhjgdgiacjfebmmmci?hl=en)

__Give me CRX__ is only used in your local browser (not the Docker Node Chrome) to obtain the .crx file of the first extension, __Proxy Auto Auth__. 

I installed on my local browser the extension __Proxy Auto Auth__, only to play with it and see what is the URL for the options of this extension.

Anyway, it seems, that the options object that you pass to Selenium, for Chrome, can take the following structure:
``` javascript
 var options = {

	host                         : '127.1.2.3',
	port                         : 4444,

	desiredCapabilities          : {
		seleniumProtocol         : 'WebDriver',
		platform                 : 'any',
		browserName              : 'chrome',
		acceptSslCerts           : true,
		cssSelectorsEnabled      : true,
		handlesAlerts            : true,
		unexpectedAlertBEhaviour : 'accept',

		proxy                    : {
			proxyType            : 'MANUAL',
			httpProxy            : 'proxycorp.com:8080'
//or you can put here user:password@proxycorp.com:8080, but not every proxy is recognized by Chrome,
//when I used my corporate proxy, Chrome is asking for the user and password and ignore what I put here
			sslProxy             : 'proxycorp.com:8080',
			ftpProxy             : 'proxyCorp.com:8080'
		},

[...]

PR #699, opened 2018-03-16
Chrome proxy support

ageekymonk:

Added support for providing proxy to the node chrome.

### DIFF
--- a/NodeChrome/README.md
+++ b/NodeChrome/README.md
@@ -20,6 +20,12 @@ Once the hub is up and running will want to launch nodes that can run tests. You
 $ docker run -d --link selenium-hub:hub -v /dev/shm:/dev/shm selenium/node-chrome
 ```
 
+If you are behind a corporate proxy, you can run it with a proxy
+
+```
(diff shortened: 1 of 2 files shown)


diemol:
Hi @ageekymonk,

Thanks for submitting this PR. Nevertheless, could you please provide more context? Isn't it just necessary to add the `-e proxy_host=proxy.example.com -e proxy_port=8080 -e no_proxy=localhost,*.dev.example.com` to the `docker run...` command? Why is it necessary to change the Chrome start command?

ageekymonk:
Thanks @diemol. In enterprise, there are proxy to reach external websites, so we need proxy settings to be in chrome to test.

diemol:
I understand the issue with proxies used in companies, I just would like to understand why the flags `-e proxy_host=proxy.example.com -e proxy_port=8080 -e no_proxy=localhost,*.dev.example.com` are not enough.

ageekymonk:
@diemol sorry, i did not understand you earlier. google chrome does not pick the proxy from the environment, it needs to be passed explicitly. 
This has been reported earlier as well https://github.com/SeleniumHQ/docker-selenium/issues/78 and it is still the same.

diemol:
I checked again and I can see that Chrome using the proxy env var, please follow this [comment](https://github.com/SeleniumHQ/docker-selenium/issues/537#issuecomment-414789754).

PR #713, opened 2018-03-26
Always download latest stable Firefox by default

testphreak:

### DIFF
--- a/NodeFirefox/Dockerfile.txt
+++ b/NodeFirefox/Dockerfile.txt
@@ -4,7 +4,7 @@ USER root
 # Firefox
 #=========
 ARG FIREFOX_VERSION=latest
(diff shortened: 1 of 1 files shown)


testphreak:
Adding a minor change, so whenever a new image is created with a new version of Selenium, whatever the latest stable Firefox is at that time, gets downloaded by default as part of the build, just as we do today with Chrome. Hopefully if this change is accepted, we will not need new pull requests for new versions of stable Firefox.

testphreak:
@diemol, @ddavison can we get this merged?

diemol:
Hi @testphreak,

Thanks for the PR! We'll do a review during the weekend to merge and then release with the new FF version 59.0.2

testphreak:
Thanks @diemol. I made a minor change to not add an else if condition to an already long line. The downside to that is folks who were previously specifying just `devedition` or `nightly`, now need to add `devedition-latest` or `nightly-latest`.

diemol:
Just tested it and it works! Thanks @testphreak

ddavison:
we already do this for Chrome, so makes sense we have this for FF.  👍

Issue #330, opened 2016-11-03
use tini in docker entry points


shankarkc:
## Meta -
Image(s):  

Both
Docker Version:  

All
## Expected Behavior -
Use tini to handle crashed defunct process. (https://github.com/krallin/tini)
## Actual Behavior -
Docker image should be able to handle zombie containers. It has been observed that many times browser crash result in defunct process and containers become zombie. Docker Jenkins uses it to avoid zombie containers/process. While running our tests we have encountered zombie containers. After using tini the stopped appearing.

diemol:
This is actually a good idea @shankarkc, perhaps you can consider to submit a PR with a proposed change?

ddavison:
super idea 👍   since no progress has been made on this, i'll go ahead and start with the change

ddavison:
note:  

> NOTE: If you are using Docker 1.13 or greater, Tini is included in Docker itself. This includes all versions of Docker CE. To enable Tini, just pass the --init flag to docker run.

[source](https://github.com/krallin/tini#using-tini)

that's nice!  let's make sure to document that

lcnja:
Let me try
[#734](https://github.com/SeleniumHQ/docker-selenium/pull/734)

diemol:
A proof of concept shows that `tini` is already included in docker since 1.13 and working properly:

Without `--init` (process ID 1 is the entry point):
```
docker run -d -p 4444:4444 --name selenium-hub selenium/hub:3.12.0-americium

docker exec -ti selenium-hub bash

seluser@b2e764c4f3a4:/$ ps aux | cat
USER       PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
seluser      1  0.1  0.0  18036  2932 ?        Ss   09:59   0:00 /bin/bash /opt/bin/entry_point.sh
seluser     10  6.0  0.8 5154544 69404 ?       Sl   09:59   0:01 java -jar /opt/selenium/selenium-server-standalone.jar -role hub -hubConfig /opt/selenium/config.json
seluser     35  0.1  0.0  18236  3216 pts/0    Ss   10:00   0:00 bash
seluser     44  0.0  0.0  34420  2804 pts/0    R+   10:00   0:00 ps aux
seluser     45  0.0  0.0   4528   664 pts/0    S+   10:00   0:00 cat

``` 

Using `--init` (process ID 1 is `/dev/init`):

```
[...]

PR #734, opened 2018-05-16
Reaping zombies and performing signal forwarding

lcnja:

Fixes: [#330](https://github.com/SeleniumHQ/docker-selenium/issues/330)


diemol:
Hi @lcnja,

Thanks for contributing! Nevertheless, two things:
* The build fails now, please have a look.
* More importantly, thanks to this [comment](https://github.com/SeleniumHQ/docker-selenium/issues/330#issuecomment-383981223) we know that tini is already included in docker. The pending task is to document how to use it with the docker-selenium images.

diemol:
@lcnja 

To be more clear, according to the [documentation](https://github.com/krallin/tini#using-tini), there is no need to install Tini. If you want, you could contribute with a PR documenting how to use it with the docker-selenium images.

lcnja:
HI @diemol 
Thanks for reply

The actual is docker from 1.13 to the latest version include tini 0.13.0

In my scene,  Sometimes still produce zombies, Here are the possible reasons.
*    Older version Tini without -p parameter
*    Default command is `/dev/init --`, Without -g parameter

So, Install the latest program manually and set the parameters correctly is the better.

image

PR #737, opened 2018-05-21
Allow to change the grid hub host

kmala:
This allows users to specify the IP to which the hub can be bind to instead of default eth0 interface. This is especially useful when there is a need to restrict access to the selenium hub by binding it to `127.0.0.1`


diemol:
Thanks for your contribution @kmala!

PR #748, opened 2018-06-26
Update to selenium version 3.13

jsa34:
New release of Selenium made - update to dockerfile to reflect this.

### DIFF
--- a/Base/Dockerfile
+++ b/Base/Dockerfile
@@ -56,5 +56,5 @@ USER seluser
 #==========
 RUN  sudo mkdir -p /opt/selenium \
   && sudo chown seluser:seluser /opt/selenium \
-  && wget --no-verbose https://selenium-release.storage.googleapis.com/3.12/selenium-server-standalone-3.12.0.jar \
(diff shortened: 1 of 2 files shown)


diemol:
Thanks @jsa34!
Could you please double check that you have signed the [Contributor License Agreement](https://github.com/SeleniumHQ/docker-selenium/blob/master/CONTRIBUTING.md#contributing-code-to-selenium)? I ask since the checkbox is not marked.

diemol:
Hm, tests are not passing. Clearly it is not related to your commit, but something on the Grid might have changed and we need to update the way it is started. 
@jsa34 I can have a look in 2-3 days, but if you feel like debugging a bit, you can build the images locally and start the containers one by one and see what the problem might be.

jsa34:
I'll take a look in a bit when I'm back at a computer... Something seems to be odd as you said with the node containers at first glance

diemol:
@jsa34 I think that was not the real error. I think that empty string in the GeckoDriver we got in the previous run was due to the GitHub Api being over the quota, and then the curl returns nothing when the releases were queried.

I just ran the build again and we got the same errors as the first time, so the tests are failing. I will do some debugging and will let you know.

PR #773, opened 2018-08-24
Adding info on how to wait for the Grid to be ready

### LINKED ISSUE #691: Add HEALTHCHECK entry to Dockerfiles
## Meta
Image(s):  standalone-chrome, standalone-firefox
Docker-Selenium Image Version(s):  2.53.0
Docker Version:  1.12.5
OS: RHEL 7

## Feature to request
When I execute my E2E tests, it may happen that I do them too soon, before the Selenium server is fully initialized. On my machine, I had to wait e.g. ~1.5 seconds before I could really connect to the server. If I don't wait, I get an exception like this:

```none
    org.openqa.selenium.remote.UnreachableBrowserException: Could not start a new session. Possible causes are invalid address of the remote server or browser start-up failure.
(...)
    Driver info: driver.version: RemoteWebDriver

        Caused by:
        java.net.SocketException: Connection reset
```

So far I've been using this code to make sure I wait enough:

```groovy
	private void waitForSeleniumServer(hostPort) {
		def healthy = (1..20).any {
			try {
				def driver = new RemoteWebDriver(new URL("http://localhost:${hostPort}/wd/hub"), healthCheckCapabilities)
				driver.quit()
				true
			}
			catch (UnreachableBrowserException e) {
				Thread.sleep 500
				false
			}
		}

		if (!healthy) {
			throw new GradleException("Timed out waiting for Selenium server to initialize on port ${hostPort}.")
		}
	}
```

However, I think it would be much better to implement a HEALTHCHECK entry in the docker images, because Gradle plugins provide support for waiting until the image is healthy. (Unfortunately, the solution may be different from what I shown above, because multiple parallel sessions might not be supported. I don't know the details of this, so I might be wrong.)

diemol:
- Fixes #691 by documenting how to add a health check when starting the containers. We won't add a health check into the docker images since it is not possible to configure the intervals for it, and we don't have data to choose values that work for everyone.
- Improved templates for new issues and PRs.


Base/check-grid.sh
```diff
@@ -0,0 +1,26 @@
...
+# check-grid.sh
+
+set -e
+
+# process arguments
+while [[ $# -gt 0 ]]
+do
+    case "$1" in
+        --host)
+        HOST="$2"
```
tnguyen14: Could you do `HOST="${2:-localhost}"`?

diemol: Good idea, I was just copying and pasting from other script. Just made the changes and I will push them now. Thanks @tnguyen14

README.md
```diff
@@ -269,6 +269,101 @@ $ FF=$(docker run --rm --name=fx \
...
+        | jq -r '.value.ready' 2>&1 | grep "true" >/dev/null; do
+    echo 'Waiting for the Grid'
+    sleep 1
+done
+
+>&2 echo "Selenium Grid is up - executing tests"
+exec $cmd
+```
+**Note:** If needed, replace `localhost` and `4444` for the correct values in your environment. Also, this script is polling indefinitely, you might want
+to teak it and establish a timeout.
```
cgoldberg: s/teak/tweak/

diemol: Thanks @cgoldberg

matt-l-w:
@diemol - sorry just pulled the code and all looks good.  I was going off of the dockerfile preview on https://hub.docker.com/r/selenium/base/~/dockerfile/ (this is still showing old code)

diemol:
No worries :)

Issue #785, opened 2018-09-12
Sporadic failure of node-firefox with "Error: cannot open display: :99"


anto-ac:
## Meta -

Image(s):  
node-firefox

Docker-Selenium Image Version(s):  
3.14.0-curium and previous versions

Docker Version:  
17.06.2-ce

OS: 
Debian 3.16.43-2 (2017-04-30)

Setup:

- wdio grid with a hub, 8 chrome nodes and 8 firefox nodes.
- Tests run on a Jenkins slave
- Here you can find a simple repository with a simplified version of the test setup: https://github.com/anto-ac/node-firefox-cannot-open-display

## Expected Behavior -

All nodes work as expected throughout the duration of the test run.

## Actual Behavior -

Sporadically, one or firefox nodes fail to have a session initiated, leading to the spec assigned to that session not being executed and the whole.
This happens mostly when there are two jobs running in parallel on the same node. 
The error is:

```
selenium-firefox-6_1  | 11:07:32.919 INFO [ActiveSessionFactory.apply] - Capabilities are: {
selenium-firefox-6_1  |   "browserName": "firefox",
selenium-firefox-6_1  |   "handlesAlerts": true,
selenium-firefox-6_1  |   "javascriptEnabled": true,
selenium-firefox-6_1  |   "locationContextEnabled": true,
selenium-firefox-6_1  |   "loggingPrefs": {
selenium-firefox-6_1  |     "browser": "ALL",
selenium-firefox-6_1  |     "driver": "ALL"
selenium-firefox-6_1  |   },
selenium-firefox-6_1  |   "maxInstances": 8,
selenium-firefox-6_1  |   "moz:firefoxOptions": {
selenium-firefox-6_1  |     "log": {
selenium-firefox-6_1  |       "level": "trace"
selenium-firefox-6_1  |     }
selenium-firefox-6_1  |   },
selenium-firefox-6_1  |   "requestOrigins": {
selenium-firefox-6_1  |     "url": "http:\u002f\u002fwebdriver.io",
selenium-firefox-6_1  |     "version": "4.13.1",
selenium-firefox-6_1  |     "name": "webdriverio"
selenium-firefox-6_1  |   },
selenium-firefox-6_1  |   "rotatable": true
selenium-firefox-6_1  | }
[...]

PR #786, opened 2018-09-13
Start from a random number when xvfb looks for available display (#785)

anto-ac:

### DIFF
--- a/NodeBase/Dockerfile.txt
+++ b/NodeBase/Dockerfile.txt
@@ -72,7 +72,9 @@ USER seluser
 #==============================
 # Scripts to run Selenium Node
 #==============================
-COPY entry_point.sh  /opt/bin/
+COPY entry_point.sh \
+  functions.sh \
+    /opt/bin/
 
 #============================
 # Some configuration options
--- a/NodeBase/entry_point.sh
+++ b/NodeBase/entry_point.sh
@@ -1,5 +1,6 @@
 #!/bin/bash
 
+source /opt/bin/functions.sh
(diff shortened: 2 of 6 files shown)


ddavison:
instead of generating a random server number, can we just use [--auto-servernum](http://manpages.ubuntu.com/manpages/trusty/man1/xvfb-run.1.html) ?

>  Try to get a free server number, starting at 99, or the argument to --server-num.

this would save us from maintaining another shell file that gets copied into all containers.

anto-ac:
We’re doing that already. It was introduced my previous PR. -a is equivalent to —auto-servernum.

We thought there could be a race condition with some conflict in the usage of the same port on different containers.

Overall we’re unsure as to why this is happening. If someone had any idea, perhaps some deeper knowledge about how xvfb works, it would be really appreciated.

anto-ac:
As you posted above, the -a options works this way:

-a --auto-servernum          
try to get a free server number, starting at --server-num

So they can be used together. You ask xvfb-run to pick an available server number using -a and you specify from which number it should start from using —server-num. If you don’t specify it, it would start from 99.

The idea behind the change is that by starting to look for a server number from a random number, the likelihood of a race condition would be much lower.

Sadly though, this change does not fix the issue itself. So while the change is harmless, it’s probably unnecessary.

It is really annoying because it makes our builds flakey.

diemol:
I'll close this PR since we saw that this does not fix the issue in your environment. We need to find out what is happening to the displays when it runs on Ubuntu.

PR #602, opened 2017-10-31
Set execution permission on entry_point.sh

eitany:
- [X] By placing an `X` in the preceding checkbox, I verify that I have signed the [Contributor License Agreement](https://github.com/SeleniumHQ/docker-selenium/blob/master/CONTRIBUTING.md#contributing-code-to-selenium)

Fix for: https://github.com/SeleniumHQ/docker-selenium/issues/601

PR #477, opened 2017-05-02
Run all processes with seluser instead of root

diemol:
While checking #474, I found that in some images the processes are run by `seluser` and in other ones they are run with `seluser` and `root`  (specially in the debug ones). E.g.:

```
docker run --rm -p 4444:4444 -p 5900:5900 -v /dev/shm:/dev/shm --shm-size=1g selenium/standalone-chrome-debug:3.4.0-chromium

docker exec -ti objective_stonebraker bash

root@be3eb04ea476:/# ps -aux
USER       PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root         1  0.0  0.1  20996  3280 ?        Ss+  14:44   0:00 /bin/bash /opt/bin/entry_point.sh
root        21  0.0  0.0  20992  1616 ?        S+   14:44   0:00 /bin/bash /opt/bin/entry_point.sh
root        26  0.0  0.0   4508  1632 ?        S+   14:44   0:00 /bin/sh /usr/bin/xvfb-run -n 99 --server-args=-screen 0 1360x1020x24 -ac +extension RANDR java -jar /opt/selenium/se
root        38  0.8  2.6 217096 54464 ?        Sl+  14:44   0:00 Xvfb :99 -screen 0 1360x1020x24 -ac +extension RANDR -nolisten tcp -auth /tmp/xvfb-run.l7EYS6/Xauthority
root        43  3.0  2.3 2966204 48284 ?       Sl+  14:44   0:01 java -jar /opt/selenium/selenium-server-standalone.jar
root        56  1.8  0.5 114212 10972 ?        S+   14:44   0:00 fluxbox -display :99.0
root        57  0.2  0.6 113380 13536 ?        S+   14:44   0:00 x11vnc -forever -usepw -shared -rfbport 5900 -display :99.0
root        78  1.0  0.1  21196  3672 ?        Ss   14:45   0:00 bash
root        87  0.0  0.1  37364  3276 ?        R+   14:45   0:00 ps -aux
```

I went through all the `Dockerfiles` and `entry_point.sh` and improved them so all the processes run now with `seluser`.

I executed the tests and they passed, I also tested it so the fix done in #459 and #469 don't get lost.

@ddavison 
I will add some comments next to the commits to make this PR easier to review.

@elgalu, can you please also help to review this PR?  

[...]

PR #787, opened 2018-09-13
Support Arbitrary User IDs for OpenShift

torstenwalter:
To quote from the official [OpenShift documentation](https://docs.openshift.com/container-platform/3.9/creating_images/guidelines.html):

> By default, OpenShift Container Platform runs containers using an arbitrarily assigned user ID. This provides additional security against processes escaping the container due to a container engine vulnerability and thereby achieving escalated permissions on the host node.

> For an image to support running as an arbitrary user, directories and files that may be written to by processes in the image should be owned by the root group and be read/writable by that group. Files to be executed should also have group execute permissions.

With this change it should be possible to run the Selenium images on OpenShift without further modifications.

### DIFF
--- a/Base/Dockerfile
+++ b/Base/Dockerfile
@@ -47,6 +47,7 @@ RUN useradd seluser \
   && usermod -a -G sudo seluser \
   && echo 'ALL ALL = (ALL) NOPASSWD: ALL' >> /etc/sudoers \
   && echo 'seluser:secret' | chpasswd
+ENV HOME=/home/seluser
 
 #=======================================
 # Create shared / common bin directory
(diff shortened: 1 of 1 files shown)


ddavison:
related:

- https://github.com/SeleniumHQ/docker-selenium/pull/602
- https://github.com/SeleniumHQ/docker-selenium/pull/502
- https://github.com/SeleniumHQ/docker-selenium/pull/477
- https://github.com/SeleniumHQ/docker-selenium/pull/185

@torstenwalter , were these changes mentioned in the four PR's above not enough for OpenShift?  It's been an ongoing battle 😜

torstenwalter:
Copared to #185 it allows to should allow users to change the config at run time.

torstenwalter:
Without the permissions change for $HOME () there will be different kinds of errors when a process tries to write/read from that directory within OpenShift. e.g.:

`Couldn't read password file: /home/seluser/.vnc/passwd`. Chrome tries to write data there and I suspect Firefox as well.

diemol:
I think this makes sense, we had to do something [similar](https://github.com/zalando/zalenium/blob/master/docker/Dockerfile#L395) to have things working in OpenShift for some users of it.

diemol:
Thanks for your contribution @torstenwalter!

PR #821, opened 2018-10-24
$GRID_DEBUG support for nodes

wildloop:
Could you please add $GRID_DEBUG support for nodes config.json?

### DIFF
--- a/NodeDebug/Dockerfile.txt
+++ b/NodeDebug/Dockerfile.txt
@@ -1,5 +1,8 @@
 USER root
 
+# Debug
+ENV GRID_DEBUG true
+
 #=====
 # VNC
 #=====
--- a/NodeBase/Dockerfile.txt
+++ b/NodeBase/Dockerfile.txt
@@ -112,6 +112,8 @@ ENV NODE_UNREGISTER_IF_STILL_DOWN_AFTER 60000
 ENV NODE_DOWN_POLLING_LIMIT 2
 # As string, maps to "applicationName"
 ENV NODE_APPLICATION_NAME ""
+# Debug
+ENV GRID_DEBUG false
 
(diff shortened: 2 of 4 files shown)


NodeDebug/Dockerfile.txt
```diff
@@ -1,5 +1,8 @@
 USER root
 
+# Debug
+ENV GRID_DEBUG true
```
diemol: Debug in the images means that people can use VNC to see what is happening inside. Some people actually use this images for normal test runs. Please do not set this env var to true for the debug images (in the near future we will merge the images anyway).

diemol:
Thanks for the PD @wildloop!
Would be nice to have some comments about the env var in the README file, with that other users would know that it is there.

NodeDebug/Dockerfile.txt
wildloop: So the meaning is closer to `inspection` than `debug`. I'll correct it, thanks.

wildloop:
@diemol - done.

diemol:
Sorry, I forgot to mention... there is a template when the PR is created. Could you please mark the checkbox after you have signed the agreement?

wildloop:
>  there is a template when the PR is created. Could you please mark the checkbox after you have signed the agreement?
I don't see any template or checkbox... where can I find them?

diemol:
Well, probably because you deleted the template. 

When you create a PR it should look like this:

image
image

wildloop:
- [X] By placing an `X` in the preceding checkbox, I verify that I have signed the [Contributor License
Agreement](https://github.com/SeleniumHQ/docker-selenium/blob/master/CONTRIBUTING.md#contnbuting-code-to-selenium)

PR #833, opened 2018-11-07
Added openssh-client (scp support)

wildloop:
What about adding these developer tools to containers:
- tcpdump - to debugging TCP connections
- mc - visual file manager
- spc - to transfer files
?

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.md
+++ b/.github/ISSUE_TEMPLATE/bug.md
@@ -41,7 +41,7 @@ Issues without a reproduction script are likely to stall and eventually be close
 ## Environment
 
 OS: <!-- Windows 10? OSX? -->
-Docker-Selenium image version: <!-- 3, 3.4, 3.141.0-actinium etc
+Docker-Selenium image version: <!-- 3, 3.4, 3.141.5-astatine etc
 Also provide the docker image id 
 -->
 Docker version: 
--- a/.github/ISSUE_TEMPLATE/bug.md
+++ b/.github/ISSUE_TEMPLATE/bug.md
@@ -41,7 +41,7 @@ Issues without a reproduction script are likely to stall and eventually be close
 ## Environment
 
 OS: <!-- Windows 10? OSX? -->
-Docker-Selenium image version: <!-- 3, 3.4, 3.14.0-krypton etc
+Docker-Selenium image version: <!-- 3, 3.4, 3.141.0-actinium etc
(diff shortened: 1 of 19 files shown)


diemol:
Hi @wildloop,

Not sure, maybe it is a good idea. But before deciding:
- Do you know how much space are the images using by adding those packages?
- What are the use cases for this packages? So we understand if it makes sense to add them in the images or if it is better just to add guidelines on how to add them in case people want to use them.

In addition, this is (I think) the first time this is suggested, so not sure if this is a common use case. In general, we prefer to have the images as simple as possible, without packages that are not needed for normal test execution (as people mostly use the images in CI pipelines where those tools would not make a difference). Maybe a good WiKi showing how to do this is better.

What do you think?

diemol:
I don't mean to be annoying or something like that, but we would like to understand better the need of including these packages...

>What are the use cases for this packages? So we understand if it makes sense to add them in the images or if it is better just to add guidelines on how to add them in case people want to use them.

In addition, if after checking your reply, we add the package, we also need docs explaining why the package is there, its use cases and examples. Would be cool to have those in the PR as well.

wildloop:
`openssh-client` is required to connect to this containers using the SCP protocol, for example to copy log file using WinSCP.

diemol:
Can't you do the same with `docker cp`?
If you want to go inside the container, you can also just do `docker exec -it <containerName> bash`, right?

wildloop:
You're right.
Everything can be done with a hammer, but it is sometimes more convenient to use pliers - so SCP vs 4MB?

barancev:
Sorry, I too can't see why `scp` is better than `docker cp`. It's not about size, it's other sort of economy.

wildloop:
Because docker cp is not a protocol that can work with [WinSCP](https://winscp.net) :)

barancev:
And why should we care about WinSCP?

barancev:
By the way, to be able to use WinSCP we should add to container an ssh server, not scp (that is a client), right?

wildloop:
To use SCP protocol the scp binary has to be present on the server - `apt install openssh-client`.
SCP is convenience because there are a lot of SCP clients: AbsoluteTelnet, Dropbear, lsh, OpenSSH, PACManager, PuTTY, SecureCRT, SmartFTP, Tera Term, WinSCP, Xshell, ZOC.

Please feel free to close this PR :)

PR #870, opened 2019-02-06
Preserve the test return code for non-travis environments

tnguyen14:
Before, even if the test fails, in non-Travis environment, the test script still return 0 because of `deactivate`. This preserves the return code from the test step so it could be returned correctly.

/cc @diemol


diemol:
Thanks for your contribution @tnguyen14!

PR #874, opened 2019-02-14
Allow proxy env variables to be set/inherited

tnguyen14:
When run in an environment that's behind a firewall, a proxy is sometimes needed to reach the Internet. I find this fix to be necessary for the tests to pass in our environment. I'm not sure if it's an appropriate change, but I figured I should at least submit.

/cc @diemol


diemol:
I think it is valid and it does not hurt to have that. Thanks @tnguyen14!

PR #883, opened 2019-03-09
Just make the version tag consistent with each other

chenrui333:
This PR is trying to reduce the confusion of the version tags in the Github issue templates. 

`3`, `3.141`, `3.141.59-krypton` are pointing to the same docker image. 

I was even thinking to include `3.141.59`, let me know if we want to pursue that.


diemol:
Thanks for your attention to detail @chenrui333!

Issue #885, opened 2019-03-19
Replace current hardcoded chromedriver version with dynamic one


mkrei:
## 🚀 Feature Proposal

Currently the Dockerfile for NodeChrome has a hardcoded version of the Chromedriver it is supposed to install, which is not too nice from a maintenance side.
With the new release system Google implemented for the Chromedriver, there should always matching versions of both Chrome and Chromedriver. As such it might make sense to switch to a dynamic install, based on the previously installed Chrome.

## Motivation

Avoid having to wait for new PRs to fix possible dependency issues caused by an update of google-chrome-stable and reduce maintenance overhead in total.

## Example

The following code (please excuse my rather crude bash code) should retrieve the best match for a given chrome version, so it would automatically update once a build is executed with a new chrome release

```bash
# get current chrome version
chrome_version=$(google-chrome --version)
version_string=$(echo "$chrome_version" | grep -oP "\d+\.\d+\.\d+\.\d+")

# extract the major version
mayor_version=$(echo "${version_string%%.*}")

# make use of the LAST_RELEASE file provided by google to get the last chromedriver for that release
wget "https://chromedriver.storage.googleapis.com/LATEST_RELEASE_${mayor_version}"
latest_version=$(cat "LATEST_RELEASE_${mayor_version}")
rm -f "LATEST_RELEASE_${mayor_version}"

# retrieve the matching chromedriver.zip
wget "https://chromedriver.storage.googleapis.com/${latest_version}/chromedriver_linux64.zip"
```
I'm not sure if this can be executed cleanly directly from the dockerfile, but it should be easy enough to put it in a .sh, execute that from the dockerfile instead of the current hardcoded download and continue with the existing code after the zip was retrieved.

diemol:
Hi @mkrei,
Sounds good, would you like to send a PR?

mkrei:
Sure, I'll give it a try. Just wanted to make sure you guys are interested in doing it as proposed.

cgoldberg:
I wonder how simultaneous the release of Chrome and Chromedriver actually are .... and if we might hit the state when one is released but not the other.

mkrei:
As long as we are talking about the stable version of Chrome, which is installed in the Dockerfile, that should not be a problem, as there are already matching builds from the beta and nightly versions. So there will already be an entry (and therefore also build) for the LAST_RELEASE_* file, which enables the retrieval of the Chromedriver.

[...]

PR #886, opened 2019-03-19
Derive Chromedriver version to install from used Chrome (#885)

mkrei:
This should replace the currently hardcoded chromedriver version string by making use of Googles new versioning scheme as well as the LASTEST_RELEASE file they provide for each major version of Chrome and Chromedriver, allowing for a match between those two.


diemol:
Thanks for your contribution @mkrei!
There was just a new Chrome release, so this comes in handy.

Issue #901, opened 2019-04-26
Dynamic chromedriver detection breaks chrome unstable


neben:
## 🐛 Bug Report

The recently introduced patch for dynamic detection of the chromedriver (75003877d98dba754d6abc66d81be0875851696e) breaks chromedriver for chrome unstable. The reason for this is that often there is no corresponding chromedriver version for chrome unstable, e.g., at the moment chrome unstable is 75, whereas the latest chromedriver version is 74. Unfortunately, the patch always tries to download the calculated chromedriver, even if the corresponding `ARG` is overridden. This breaks the container build.

## To Reproduce

```
# docker build --no-cache https://github.com/SeleniumHQ/docker-selenium.git#3.141.59-neon:NodeChrome --build-arg CHROME_VERSION=google-chrome-unstable --build-arg CHROME_DRIVER_VERSION=LATEST_RELEASE_74
Sending build context to Docker daemon  17.41kB
Step 1/12 : FROM selenium/node-base:3.141.59-neon
 ---> c0255bd8f01b
Step 2/12 : LABEL authors=SeleniumHQ
 ---> Running in e57fafe4e52f
Removing intermediate container e57fafe4e52f
 ---> eda44811d5c6
Step 3/12 : USER root
 ---> Running in 99975cdeb0f5
Removing intermediate container 99975cdeb0f5
 ---> d214374f2844
Step 4/12 : ARG CHROME_VERSION="google-chrome-stable"
 ---> Running in d2a2ba1ef652
Removing intermediate container d2a2ba1ef652
 ---> 723fe7e391b9
Step 5/12 : RUN wget -q -O - https://dl-ssl.google.com/linux/linux_signing_key.pub | apt-key add -   && echo "deb http://dl.google.com/linux/chrome/deb/ stable main" >> /etc/apt/sources.list.d/google-chrome.list   && apt-get update -qqy   && apt-get -qqy install     ${CHROME_VERSION:-google-chrome-stable}   && rm /etc/apt/sources.list.d/google-chrome.list   && rm -rf /var/lib/apt/lists/* /var/cache/apt/*
 ---> Running in ab0ef558afbc
[...]
Setting up google-chrome-unstable (75.0.3770.9-1) ...
update-alternatives: using /usr/bin/google-chrome-unstable to provide /usr/bin/x-www-browser (x-www-browser) in auto mode
[...]

PR #903, opened 2019-04-30
don't detect chromedriver version if provided as arg (#901)

neben:
This fixes cases where there was no chromedriver available for
chrome unstable releases, resulting in failed image builds,
even if an existing chromedriver version was provided by the
user.

### DIFF
--- a/NodeChrome/Dockerfile
+++ b/NodeChrome/Dockerfile
@@ -41,24 +41,18 @@ USER seluser
 # Latest released version will be used by default
 #============================================
 ARG CHROME_DRIVER_VERSION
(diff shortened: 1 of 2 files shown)


NodeChrome/Dockerfile
```diff
@@ -41,24 +41,18 @@ USER seluser
 # Latest released version will be used by default
 #============================================
 ARG CHROME_DRIVER_VERSION
-RUN CHROME_STRING=$(google-chrome --version) \
```
diemol: Please make changes only on the `Dockerfile.txt` file.

neben: ok, i reverted the change

PR #933, opened 2019-06-20
Provide classpath instead of jar when invoking java

doublemarket:
When starting a Selenium process in a container, the startup script invokes the `java` command with the `-jar` option. However, it's not possible to add classpaths when running a Selenium process even if we use the `JAVA_OPTS` because [when the `-jar` option is used, the `-cp` option which provides classpaths is ignored](https://bugs.java.com/bugdatabase/view_bug.do?bug_id=4459663). If we want to use a custom proxy or an add-on as a Jar file without changing a container, `-cp` is needed.

This pull request makes a change to the startup scripts to use `-cp` instead of `-jar` and introduce the `JAVA_CLASSPATH` environment variable to overwrite the default classpaths.

This shouldn't break existing usages without providing the `JAVA_CLASSPATH` environment variable.

Do you think this is valuable to have? If so, I'll update the README file as well.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -41,7 +41,7 @@ Issues without a reproduction script are likely to stall and eventually be close
 ## Environment
 
 OS: <!-- Windows 10? OSX? -->
-Docker-Selenium image version: <!-- 3, 3.141, 3.141.59-palladium etc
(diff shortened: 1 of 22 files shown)


doublemarket:
@diemol Thanks for reviewing and I've updated the README file with an example!

PR #1385, opened 2021-09-08
Add configurable docker env variables to support disabling novnc and changing ports

JasonBerenbach:
### Description
 
Adds env variables to allow for configuring the novnc script, and the ports being used. 

### Motivation and Context

It may be desirable to disable the novnc script after starting xvfm service if net=host. The listening port was hardcoded and not configurable, so this adds an env variable that can be 
set to update that. Also the vnc port is already configurable so this ensures that if the vnc_port is set, it will update the value in the start_no_vnc script as well. 

### Types of changes

### Checklist

### DIFF
--- a/NodeBase/Dockerfile.txt
+++ b/NodeBase/Dockerfile.txt
@@ -160,6 +160,7 @@ ENV SCREEN_DPI 96
 ENV DISPLAY :99.0
 ENV DISPLAY_NUM 99
 ENV START_XVFB true
+ENV START_NO_VNC true
 
 #========================
 # Selenium Configuration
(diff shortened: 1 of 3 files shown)


diemol (approved):
Make sense, thank you, @JasonBerenbach!

therealdjryan:
If the NO_VNC_PORT is set, I think se:vncLocalAddress should also be set

`DriverServiceSessionFactory.java:`
``` 
private Capabilities readVncEndpoint(Capabilities requestedCaps, Capabilities returnedCaps) {
    String seVncEnabledCap = "se:vncEnabled";
    String seVncEnabled = String.valueOf(requestedCaps.getCapability(seVncEnabledCap));
    if (Boolean.parseBoolean(seVncEnabled)) {
      returnedCaps = new PersistentCapabilities(returnedCaps)
        .setCapability("se:vncLocalAddress", "ws://localhost:7900/websockify")
        .setCapability(seVncEnabledCap, true);
    }
    return returnedCaps;
  }
```

PR #1414, opened 2021-10-05
Enable deployment of selenium docker images to restrictive Kubernetes

ptriller1und1:
### Description
This patch makes it easier to use the Selenium Containers provided in kubernetes clusters with restrictive security settings.

### Motivation and Context
I want to use the Senenium Docker images in a Kubernetes Cluster which only allows deployments that set `readOnlyRootFilesystem: true` 
In General the containers support this well except for one issue. The selenium config file is hard coded to be `/opt/selenium/config.toml`. This makes mounting a writable volume at this position impossible since this directory contains rather important jar files.
My change makes  it possible to change the location of the configuration file with an environment variable and it also allows the user the ability to disable the generation of the configuration. This way an external configuration could be injected into it.

- `CONFIG_FILE`: absolute path to the configuration file inside the container. Defaults to  `/opt/selenium/config.toml`
- `GENERATE_CONFIG`: Should the selenium configuration be generated. Defaults to  `true`

My change is 100 % backward compatible and does not change any behavior unless the relevant environment Variables are set.

### Types of changes

### Checklist


diemol:
Apologies for the delay reviewing this PR, we've been busy with the 4.0 release. We will review this in the next couple of days. Thanks!

Base/entry_point.sh
```diff
@@ -11,6 +11,10 @@ if ! whoami &> /dev/null; then
   fi
 fi
 
+# create directories in case of mounted emptyDir
+mkdir -p /var/run/supervisor
+mkdir -p /var/log/supervisor
```
diemol: I don't understand this part. Is this a workaround in case someone mounts a volume to `/var/run/supervisor` and `/var/log/supervisor`?

diemol (commented):
This looks fine, I just left a comment to understand better if the lines are needed in the `entry_point.sh` file.

Base/entry_point.sh
ptriller1und1: It is true that this is not strictly necessary and it probably can be removed.  This is a call made under the assumption the the /var/run or /var/log are mounted as an emptyDir. Then the Subdirectories do not exist. This was done because I was not sure if other logfiles are written outside of the var/log/supervisor directory. Now I am pretty sure this is not the case so this could be removed.
For Example in my case I mount "emptyDir" Volumes at /tmp, /var/tmp /var/run and /var/log, and I use the new configuration option from this patch to move the config file to /tmp/selenium.toml
If we remove this mkdir the emptyDir could either be mounted to /var/run/supervisor and  /var/run/supervisor or the mkdir has to be made in the kubernetes deployment descriptor. Both is not really problematic, so I could remove this if you want.

diemol: Thanks for your reply, @ptriller1und1.

I understand this might be needed in some cases, but those cases are not the general use case, which is why I think it is better if we remove this bit from the PR. After that, I'd be happy to merge it.

ptriller1und1: I changed it in the merge request.

diemol (approved):
Thank you, @ptriller1und1!

PR #1543, opened 2022-04-07
Pin actions to a full length commit SHA

naveensrinivasan:
- Pinned actions by SHA https://github.com/ossf/scorecard/blob/main/docs/checks.md#pinned-dependencies

>Pin actions to a full length commit SHA

>Pinning an action to a full length commit SHA is currently the only way to use an action as an immutable release. Pinning to a particular SHA helps mitigate the risk of a bad actor adding a backdoor to the action's repository, as they would need to generate a SHA-1 collision for a valid Git object payload.

https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions

[How do I validate these pinned actions?](https://gist.github.com/naveensrinivasan/ca008c07279176acce28969fb77d056f)

Also, dependabot supports upgrading based on SHA. https://github.com/ossf/scorecard/pull/1700

GitHub's own repository pin's their checkout actions by SHA and doesn't use the version tag
https://github.com/github/docs/blob/ea7f218c91ecbae9a700a8702b51a7d2736e0d2c/.github/workflows/docs-review-collect.yml#L23

Signed-off-by: naveensrinivasan <172697+naveensrinivasan@users.noreply.github.com>

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -51,25 +51,25 @@ jobs:
           DOCKER_USERNAME: ${{secrets.DOCKER_USERNAME}}
           DOCKER_PASSWORD: ${{secrets.DOCKER_PASSWORD}}
       - name: Deploy new images
-        uses: nick-invision/retry@v2
+        uses: nick-invision/retry@7f8f3d9f0f62fe5925341be21c2e8314fd4f7c7c # v2
         with:
           timeout_minutes: 20
           max_attempts: 3
           command: VERSION="${GRID_VERSION}" BUILD_DATE=${BUILD_DATE} make release
       - name: Tag images as latest
-        uses: nick-invision/retry@v2
+        uses: nick-invision/retry@7f8f3d9f0f62fe5925341be21c2e8314fd4f7c7c # v2
         with:
           timeout_minutes: 20
           max_attempts: 3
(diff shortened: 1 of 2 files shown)


diemol (commented):
You mention that 
>Also, dependabot supports upgrading based on SHA. https://github.com/ossf/scorecard/pull/1700

But how will I know which tag am I using? Seems the comments in those workflows were added manually. Readability is important for us and we check the actions before adding them to our workflows.

naveensrinivasan:
> You mention that
> 
> > Also, dependabot supports upgrading based on SHA. [ossf/scorecard#1700](https://github.com/ossf/scorecard/pull/1700)
> 
> But how will I know which tag am I using? Seems the comments in those workflows were added manually. Readability is important for us and we check the actions before adding them to our workflows.

I understand that is an issue.  There isn't a straight way out now TBH. There needs to be some `grep` and `awk` of the file to parse the SHA and use something like this to relate the SHA to a tag.
https://gist.github.com/naveensrinivasan/ca008c07279176acce28969fb77d056f


The goal is to avoid something like this https://julienrenaux.fr/2019/12/20/github-actions-security-risk/

diemol (approved):
OK, let's give it a try. If it becomes hard to maintain or dependabot does not help, we will probably revert.

PR #1582, opened 2022-05-25
Trigger chart release only on changes in the charts directory

den-is:
### Motivation and Context

I would like to suggest triggering chart-releaser action only on changes in the charts directory.
I don't see any other logic that triggers chart publishing on any push to the trunk that also updates Chart.yaml with a new version automatically.
So IMHO, Chart version is updated and pushed to trunk manually (PR merged), so the chart needs release only when anything in the charts dir changes.

Feel free to discard, if that logic doesn't fit your workflow :)

### Types of changes

### Checklist

### DIFF
--- a/.github/workflows/helm-chart-release.yml
+++ b/.github/workflows/helm-chart-release.yml
@@ -4,6 +4,8 @@ on:
   push:
     branches:
       - trunk
+    paths:
+      - 'charts/**'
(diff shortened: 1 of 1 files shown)


diemol (approved):
@den-is, makes a lot of sense, thank you!

PR #1606, opened 2022-06-16
Fix typo in Readme

rklec:
It's just a typo fix so I deleted this template.


rklec:
It's a minor fix and does not require such a thing as any threshold of originality is never reached.

diemol (approved):

Issue #1688, opened 2022-10-01
[🚀 Feature]: Auto scaling Seleniun Grid running in Kubernetes using KEDA


prashanth-volvocars:
### Feature and motivation

Auto scaling selenium grid was a problem that was pending to be solved for long. So i took it up when there was a requirement at current my work place. KEDA seemed to the best candidate for the Job and i wrote a new scalar for Selenium Grid a year ago. I would like to have this enabled by default in our charts. But i am not sure how to make it a part of the current setup. I have shared my whole setup [here](https://github.com/prashanth-volvocars/docker-selenium/tree/auto-scaling).

Any suggestions on how to make this part of the default helm charts and docker images would be great. I would want this to be a part of this repo so it could benefit others.

### Usage example

You can grab all information about the setup [here](https://medium.com/@rprashanth27/auto-scaling-selenium-grid-using-keda-easy-way-out-97d7f7d19a35)

On a nutshell, It can

- Auto Scale browser nodes up and down.
- Record videos and store them named under session id
- Capture logs and store them named under session id
- Upload captured logs and videos to S3

jamesmortensen:
Hi @prashanth-volvocars It looks like you forked and made the changes already, so why don't you open a pull request. That will make it easier for us to review the changes, give feedback, and hopefully get the changes merged.

prashanth-volvocars:
Hi @jamesmortensen 

Sure i have raised a pull request https://github.com/SeleniumHQ/docker-selenium/pull/1689 with the above mentioned changes

abhitrivedi2006:
Hi @prashanth-volvocars .. I have deployed your solution on AKS. I have intentionally set Keda max chrome count to 15 Pods. expecting hub to hold the tests in the queue and once slots are available then submit jobs to them.

when looked at the logs I see below msg which I am not able to understand as where are setting up this .. also attaching full logs from Hub.. 


Unable to create session: Could not start a new session. Could not start a new session. **Max session count reached.**



[...]

PR #1689, opened 2022-10-02
Autoscaling selenium grid on kubernetes with video recording

prashanth-volvocars:
### Description
Autoscale selenium browser nodes running in kubernetes based on the request pending in session queue using KEDA. It also includes ability to automatically record videos using ffmpeg and capture network, browser logs. 
The recorded videos and logs are stored under <session_id> name and can be uploaded directly to S3.

### Motivation and Context
Auto scaling selenium grid was a problem that was pending to be solved for long. So i took it up when there was a requirement at current my work place. KEDA seemed to the best candidate for the Job and i wrote a new scalar for Selenium Grid a year ago. I would like to have this enabled by default in our charts so everyone could use it.

### Types of changes

### Checklist

### DIFF
--- a/NodeBase/Dockerfile
+++ b/NodeBase/Dockerfile
@@ -53,6 +53,15 @@ RUN apt-get update -qqy \
     fluxbox \
   && rm -rf /var/lib/apt/lists/* /var/cache/apt/*
 
+#=========
+# ffmpeg
+# A fast, lightweight and responsive window manager
+#=========
+RUN apt-get update -qqy \
+  && apt-get -qqy install \
+    ffmpeg \
+  && rm -rf /var/lib/apt/lists/* /var/cache/apt/*
+
 #================
 # Font libraries
 #================
@@ -109,6 +118,13 @@ RUN  wget -nv -O noVNC.zip \
   && rm -rf websockify-${WEBSOCKIFY_SHA}/tests \
   && mv websockify-${WEBSOCKIFY_SHA} /opt/bin/noVNC/utils/websockify
 
+#===================================================
+# Install aws cli
+#===================================================
+RUN wget -O awscliv2.zip "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip"\
+  && unzip awscliv2.zip \
+  && sudo ./aws/install
+
 #=========================================================================================================================================
 # Run this command for executable file permissions for /dev/shm when this is a "child" container running in Docker Desktop and WSL2 distro
 #=========================================================================================================================================
@@ -125,6 +141,8 @@ USER 1200
 #==============================
 COPY start-selenium-node.sh \
       start-xvfb.sh \
+      start-video-recorder.sh \
+      start-s3-uploader.sh \
       /opt/bin/
 
 #==============================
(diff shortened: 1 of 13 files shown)


jamesmortensen:
Thank you for submitting the pull request regarding issue #1688  Please note it may take us some time to find someone who can review the changes.

prashanth-volvocars:
Sure thanks @jamesmortensen for the swift response. Hope to see this getting merged. Also there is a similar issue in the selenium project https://github.com/SeleniumHQ/selenium/issues/9845 on the same topic

mhnaeem:
@prashanth-volvocars Thank you for this work, my team is looking for an auto scaling solution and this PR would be great to have.

**_Personal opinion below - please wait for reply from contributors before making any changes_**
I do have one concern and perhaps @jamesmortensen or @diemol could chime in. Is it possible to make this video upload process cloud agnostic, since a lot of users won't be using AWS S3 buckets. Perhaps there is some way to let the user decide how the videos get processed or uploaded after being captured instead of the selenium repo handling the upload process which also comes at a cost of installing aws cli in every docker image.

Maybe we can have two PRs addressing the two separate parts - Autoscaling and Video Capture.

jamesmortensen:
@mhnaeem Thanks for jumping in. It does seems like we have two separate PR's here, and I think it would be easier to evaluate them if we could see them as separate PR's.  @prashanth-volvocars what do you think?

----
Here's my feedback for the video recording portion of the PR:

Regarding the changes in NodeBase, I don't see any major issues with installing the AWS CLI. We do already install components that not everyone uses.  For instance, for those who don't use VNC or noVNC, those components, including xvfb and fluxbox, are installed in the container images and end up doing nothing.  Do we know how much extra space it takes to install the AWS CLI? If it's not trivial, then perhaps there are other ideas?

Regarding adding other cloud providers, I don't think we need to be concerned about that right now, I think we could merge with what we have.  At the same time, we could make some space for it with a few abstract changes.  For example, `start-s3-uploader.sh` could be encapsulated inside `start-uploader.sh` and then `start-uploader.sh` would either call `start-s3-uploader.sh` or `start-gcp-uploader.sh` depending on the environment variables.  This would allow other providers to be easily added without disturbing other logic. If someone wanted to upload videos to Backblaze B2, then they'd add an implementation for `start-b2-uploader.sh` and open a pull request.  We could potentially even script it in a way that the Dockerfile doesn't need to be modified for someone to add another provider.

What do you both think?

prashanth-volvocars:
Hello @jamesmortensen

Yeah sure we can have them as two separate PR's.

For the video upload to cloud provider. We can change it as per your suggestion to use start-uploader.sh and inside it call a scipt specific to a cloud provider. I will make these changes when i find some time and i will update the PR.

jamesmortensen:
Sounds great.  :)

diemol (commented):
I apologize for the delay in a proper review. In general, yes, it'd be nice to have to separate PRs. Basically because this one is adding `ffmpeg` to the Node images, and that is a huge change how we could handle video recording. 

I prefer to have a PR that only focuses on the Keda integration, and when we are there, we can discuss the video recording approach.

prashanth-volvocars:
@diemol @jamesmortensen 

Closing this in favour of https://github.com/SeleniumHQ/docker-selenium/pull/1714

PR #1704, opened 2022-10-24
Exiting container when Standalone or Node exits, even with error exit status

### LINKED ISSUE #1703: [🚀 Feature]: Exiting container when Standalone or Node exits (on error too)
### Feature and motivation

From https://github.com/SeleniumHQ/docker-selenium/pull/1504

> The behaviour we have always had is one where supervisor controls all processes and keeps the container running. This is great because it has improved maintainability and readability. However, when the most important process, the Selenium Server, exits... the container has no purpose anymore. Users need to stop/kill the container themselves.
>
> This change will make the container exit when the Selenium Server process exits.

It would be nice if the container would exit even if the Selenium Server process exits with an error code (e.g. processes is killed because of OOM).

Instead of
```
command=bash -c "/opt/bin/start-selenium-node.sh && kill -s SIGINT `cat /var/run/supervisor/supervisord.pid`"
```
we should have
```
command=bash -c "/opt/bin/start-selenium-node.sh; kill -s SIGINT `cat /var/run/supervisor/supervisord.pid`"
```

### Usage example

K8s will restart the pod if the Selenium Server process is not available anymore.

DamianFekete:
### Description
Fixes https://github.com/SeleniumHQ/docker-selenium/issues/1703

Stop supervisord if the Standalone or Node exits, even with an error exit status.

### Motivation and Context
If the selenium server process is for example killed (exit status != 0), supervisord continues to run, and Kubernetes won't restart the entire pod.

### Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)

I'll mark it as breaking change, but I hope people don't really rely on this functionality.

### Checklist

### DIFF
--- a/Standalone/selenium.conf
+++ b/Standalone/selenium.conf
@@ -61,7 +61,7 @@ stderr_capture_maxbytes=50MB
 
 [program:selenium-standalone]
 priority=15
-command=bash -c "/opt/bin/start-selenium-standalone.sh && kill -s SIGINT `cat /var/run/supervisor/supervisord.pid`"
+command=bash -c "/opt/bin/start-selenium-standalone.sh; EXIT_CODE=$?; kill -s SIGINT `cat /var/run/supervisor/supervisord.pid`; exit $EXIT_CODE"
 stopasgroup = true
 autostart=true
 autorestart=false
--- a/NodeBase/selenium.conf
+++ b/NodeBase/selenium.conf
@@ -55,7 +55,7 @@ stderr_capture_maxbytes=50MB
 
 [program:selenium-node]
 priority=15
-command=bash -c "/opt/bin/start-selenium-node.sh && kill -s SIGINT `cat /var/run/supervisor/supervisord.pid`"
+command=bash -c "/opt/bin/start-selenium-node.sh; EXIT_CODE=$?; kill -s SIGINT `cat /var/run/supervisor/supervisord.pid`; exit $EXIT_CODE"
 stopasgroup = true
 autostart=true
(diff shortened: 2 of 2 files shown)


Standalone/selenium.conf
```diff
@@ -61,7 +61,7 @@ stderr_capture_maxbytes=50MB
 
 [program:selenium-standalone]
 priority=15
-command=bash -c "/opt/bin/start-selenium-standalone.sh && kill -s SIGINT `cat /var/run/supervisor/supervisord.pid`"
+command=bash -c "/opt/bin/start-selenium-standalone.sh; EXIT_CODE=$?; kill -s SIGINT `cat /var/run/supervisor/supervisord.pid`; exit $EXIT_CODE"
```
DamianFekete, line 64: Using `exit $EXIT_CODE` to make the exit status of `start-selenium-standalone.sh` visible in logs, otherwise it would be the exit status of the kill command.
```
seluser@faf20ce83332:/etc/supervisor/conf.d$ /opt/bin/entry_point.sh
...
2022-10-24 10:40:06,323 INFO spawned: 'selenium-node' with pid 1144
2022-10-24 10:40:06,330 INFO success: selenium-node entered RUNNING state, process has stayed up for > than 0 seconds (startsecs)
2022-10-24 10:40:06,339 INFO exited: selenium-node (exit status 1; not expected)
2022-10-24 10:40:06,339 WARN received SIGINT indicating exit request
...
```

diemol (approved):
Thank you, @DamianFekete!

PR #1709, opened 2022-10-27
[grid][chart] Allows users to turn off Deployment creation for Nodes

### LINKED ISSUE #1708: [🚀 Feature]: Allow users to turn off creation for Deployment in k8s
### Feature and motivation

I would like to stop the creation of `Deployment` type in my Kubernetes cluster and instead use a custom resource type or `Job`. This feature will provide users with a flag to turn off the creation of browser node deployments.

### Usage example

I want to use `ScaledJob` by Keda instead of the default `Deployment` but I get multiple Nodes, ones from Deployments and others from Jobs which is confusing.

mhnaeem:
### Description
This change introduces a new value called `deploymentEnabled` for each of the node browsers, if the value is set to `true` Deployments will be created for each of the respective nodes however if `false` no Deployments will be created.

This will allow users to turn off creation of the default deployments and to use other resource types instead such as jobs.

### Motivation and Context

Fixes #1708

### Types of changes

- [X] New feature (non-breaking change which adds functionality)

### Checklist

- [X] I have read the [contributing](https://selenium.dev/documentation/en/contributing/) document.

### DIFF
--- a/charts/selenium-grid/CHANGELOG.md
+++ b/charts/selenium-grid/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 All notable changes to this helm chart will be documented in this file.
 
+## :heavy_check_mark: 0.13.0
+
+### Added
+- Added support to disable Chrome, Edge, and Firefox Deployment using `deploymentEnabled`
+
 ## :heavy_check_mark: 0.12.2
 
 ### Changed
--- a/charts/selenium-grid/Chart.yaml
+++ b/charts/selenium-grid/Chart.yaml
@@ -2,6 +2,6 @@ apiVersion: v2
 name: selenium-grid
 description: A Helm chart for creating a Selenium Grid Server in Kubernetes
 type: application
-version: 0.12.2
+version: 0.13.0
 appVersion: 4.5.3-20221024
 icon: https://github.com/SeleniumHQ/docker-selenium/raw/trunk/logo.png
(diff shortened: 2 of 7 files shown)


diemol (changes requested):
Can you also please update the changelog and the `Chart.yaml` file?

mhnaeem:
@diemol Can you please take a look at the `Chart.yaml` file again? I couldn't find any instructions on what update was required so I just bumped up the version and added a line in Changelog

diemol (approved):
That looks, perfect, thank you!

ghost:
Sorry for coming in so late, but I don't understand this feature. Why not just use `.Values.chromeNode.enabled` instead of introducing a new variable that does the same?

I also see that this new variable is used with the same logic as `.Values.chromeNode.enabled`.

```yaml
{{- if and .Values.chromeNode.enabled .Values.chromeNode.deploymentEnabled }}
```
Both variables are dependent on each other now? 
If someone did not want the chrome nodes to be deployed, they would just set `.Values.chromeNode.enabled = false` 

@mhnaeem can you please explain the added value here?

mhnaeem:
For our use case we needed to use Keda `ScaledObjects` and/or `ScaledJobs` in our workflow and not the default `Deployment` which kept creating extra pods for no reason. The solution was to simply disable the default deployment and use a custom one created by Keda. Even with 0 replicas this `Deployment` item existed in our cluster.

Plus I am hoping that the autoscaling PR #1714 goes out soon which will use `ScaledObjects`. We can then possibly have `scaleobjects.enabled` and `scaledjobs.enabled` values so users can even further configure what type of pods they need instead of being stuck with the default one for autoscaling.

I did not think much of this change because ` .Values.chromeNode.service.enabled` already exists and does about the same thing

PR #1714, opened 2022-11-01
Autoscaling selenium grid on kubernetes

prashanth-volvocars:
### Description
Autoscale selenium browser nodes running in kubernetes based on the request pending in session queue using KEDA. Toggle autoscaling on/off using 'autoscalingEnabled' option in helm charts.

### Motivation and Context
Auto scaling selenium grid was a problem that was pending to be solved for long. So i took it up when there was a requirement at current my work place. KEDA seemed to the best candidate for the Job and i wrote a new scalar for Selenium Grid a year ago. I would like to have this enabled by default in our charts so everyone could use it.

### Types of changes

### Checklist

### DIFF
--- a/charts/selenium-grid/README.md
+++ b/charts/selenium-grid/README.md
@@ -29,6 +29,11 @@ helm install selenium-grid docker-selenium/selenium-grid --version <version>
 helm install selenium-grid --set ingress.hostname=selenium-grid.k8s.local docker-selenium/chart/selenium-grid/.
 ```
 
+## Enable Selenium Grid Autoscaling
+Selenium Grid has the ability to autoscale browser nodes up/down based on the requests pending in session queue. You can enable it setting 'autoscalingEnabled' to `true`. You need to install KEDA by following the [instructions](https://keda.sh/docs/2.8/deploy/#helm) in order for autoscaling to work. 
+
+The hpa.url value is configured to work for grid installed in `default` namespace. If you are installing the grid in some other namespace make sure to update the value of hpa.url accordingly. 
+
 ## Updating Selenium-Grid release
 
 Once you have a new chart version, you can update your selenium-grid running:
@@ -66,6 +71,7 @@ This table contains the configuration parameters of the chart and their default 
 | `ingress.annotations`                   | `{}`                               | Custom annotations for ingress resource                                                                                    |
 | `ingress.hostname`                      | `selenium-grid.local`              | Default host for the ingress resource                                                                                      |
 | `ingress.tls`                           | `[]`                               | TLS backend configuration for ingress resource                                                                             |
+| `ingress.path`                          | `/`                                | Default path for ingress resource                                                                             |
 | `busConfigMap.annotations`              | `{}`                               | Custom annotations for configmap                                                                                           |
 | `chromeNode.enabled`                    | `true`                             | Enable chrome nodes                                                                                                        |
 | `chromeNode.deploymentEnabled`          | `true`                             | Enable creation of Deployment for chrome nodes                                                                             |
@@ -95,6 +101,11 @@ This table contains the configuration parameters of the chart and their default 
 | `chromeNode.lifecycle`                  | `{}`                               | hooks to make pod correctly shutdown or started                                                                            |
 | `chromeNode.extraVolumeMounts`          | `[]`                               | Extra mounts of declared ExtraVolumes into pod                                                                             |
 | `chromeNode.extraVolumes`               | `[]`                               | Extra Volumes declarations to be used in the pod (can be any supported volume type: ConfigMap, Secret, PVC, NFS, etc.)     |
+| `chromeNode.autoscalingEnabled`         | `false`                            | Enable/Disable autoscaling of browser nodes |
+| `chromeNode.hpa.url`                    | `http://selenium-hub.default:4444/graphql` | Graphql Url of the hub or the router |
+| `chromeNode.hpa.browserName`            | `chrome`                           | BrowserName from the capability |
+| `chromeNode.hpa.browserVersion`         | ``                                 | BrowserVersion from the capability |
+| `chromeNode.maxReplicaCount`            | `8`                                | Max number of replicas that this browsernode can auto scale up to |
 | `firefoxNode.enabled`                   | `true`                             | Enable firefox nodes                                                                                                       |
 | `firefoxNode.deploymentEnabled`         | `true`                             | Enable creation of Deployment for firefox nodes                                                                            |
 | `firefoxNode.replicas`                  | `1`                                | Number of firefox nodes                                                                                                    |
@@ -123,6 +134,11 @@ This table contains the configuration parameters of the chart and their default 
 | `firefoxNode.lifecycle`                 | `{}`                               | hooks to make pod correctly shutdown or started                                                                            |
 | `firefoxNode.extraVolumeMounts`         | `[]`                               | Extra mounts of declared ExtraVolumes into pod                                                                             |
 | `firefoxNode.extraVolumes`              | `[]`                               | Extra Volumes declarations to be used in the pod (can be any supported volume type: ConfigMap, Secret, PVC, NFS, etc.)     |
+| `firefoxNode.autoscalingEnabled`        | `false`                            | Enable/Disable autoscaling of browser nodes |
+| `firefoxNode.hpa.url`                   | `http://selenium-hub.default:4444/graphql` | Graphql Url of the hub or the router |
+| `firefoxNode.hpa.browserName`           | `firefox`                          | BrowserName from the capability |
+| `firefoxNode.hpa.browserVersion`        | ``                                 | BrowserVersion from the capability |
+| `firefoxNode.maxReplicaCount`           | `8`                                | Max number of replicas that this browsernode can auto scale up to |
 | `edgeNode.enabled`                      | `true`                             | Enable edge nodes                                                                                                          |
 | `edgeNode.deploymentEnabled`            | `true`                             | Enable creation of Deployment for edge nodes                                                                               |
 | `edgeNode.replicas`                     | `1`                                | Number of edge nodes                                                                                                       |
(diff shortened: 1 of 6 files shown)


prashanth-volvocars:
@diemol @jamesmortensen As discussed earlier, I have raised two separate PR's for autoscaling and video recoriding.

5ryde (approved):

jamesmortensen:
Hi @prashanth-volvocars thank you for splitting them up.  We'll take a look as soon as we can.

charts/selenium-grid/README.md
```diff
@@ -29,6 +29,11 @@ helm install selenium-grid docker-selenium/selenium-grid --version <version>
 helm install selenium-grid --set ingress.hostname=selenium-grid.k8s.local docker-selenium/chart/selenium-grid/.
 ```
 
+## Enable Selenium Grid Autoscaling
+Selenium Grid has the ability to autoscale browser nodes up/down based on the requests pending in session queue. You can enable it setting 'autoscalingEnabled' to `true`. You need to install KEDA by following the [instructions](https://keda.sh/docs/2.8/deploy/#helm) in order for autoscaling to work.
```
diemol:
```suggestion
Selenium Grid has the ability to autoscale browser nodes up/down based on the pending requests in the 
session queue. 
You can enable it by setting 'autoscalingEnabled' to `true`. You need to install KEDA by following the 
[instructions](https://keda.sh/docs/2.8/deploy/#helm) in order for autoscaling to work. 
```

charts/selenium-grid/README.md
```diff
@@ -29,6 +29,11 @@ helm install selenium-grid docker-selenium/selenium-grid --version <version>
 helm install selenium-grid --set ingress.hostname=selenium-grid.k8s.local docker-selenium/chart/selenium-grid/.
 ```
 
+## Enable Selenium Grid Autoscaling
+Selenium Grid has the ability to autoscale browser nodes up/down based on the requests pending in session queue. You can enable it setting 'autoscalingEnabled' to `true`. You need to install KEDA by following the [instructions](https://keda.sh/docs/2.8/deploy/#helm) in order for autoscaling to work. 
+
+The hpa.url value is configured to work for grid installed in `default` namespace. If you are installing the grid in some other namespace make sure to update the value of hpa.url accordingly.
```
diemol, line 39:
```suggestion
The `hpa.url` value is configured to work for Grid when installed in the `default` namespace. If you are installing
the Grid in some other namespace make sure to update the value of `hpa.url` accordingly. 
```

charts/selenium-grid/values.yaml
```diff
@@ -363,7 +364,7 @@ chromeNode:
     # Custom annotations for service
     annotations: {}
   # Size limit for DSH volume mounted in container (if not set, default is "1Gi")
-  dshmVolumeSizeLimit: 1Gi
+  dshmVolumeSizeLimit: 2Gi
```
diemol: Why? This would affect people who are already using the chart, their clusters have been provisioned with the existing values.

charts/selenium-grid/values.yaml
```diff
@@ -375,17 +376,17 @@ chromeNode:
     # failureThreshold: 120
     # periodSeconds: 5
   # Time to wait for pod termination
-  terminationGracePeriodSeconds: 30
+  terminationGracePeriodSeconds: 3600
```
diemol: Why?

charts/selenium-grid/values.yaml
```diff
@@ -375,17 +376,17 @@ chromeNode:
...
-    #         while curl 127.0.0.1:5555/status; do sleep 1; done
+  lifecycle: 
+    preStop:
+      exec:
+        command:
+          - bash
+          - -c
+          - |
+            curl -X POST 127.0.0.1:5555/se/grid/node/drain --header 'X-REGISTRATION-SECRET;' && \
+            while curl 127.0.0.1:5555/status; do sleep 1; done;
```
diemol, line 389: Shouldn't this `preStop` be enabled only if `autoscalingEnabled` is `true`?

charts/selenium-grid/values.yaml
```diff
@@ -398,6 +399,15 @@ chromeNode:
   #   persistentVolumeClaim:
   #     claimName: my-pv-claim
 
+  # Keda scaled object configuration
+  autoscalingEnabled: false
+  maxReplicaCount: 8
+  hpa:
+    url: http://selenium-hub.default:4444/graphql # Repalce your http graphql url here
```
diemol, line 406: `selenium-hub.default` can this be retrieved from the Hub or Router service or pod name?

charts/selenium-grid/values.yaml
```diff
@@ -475,7 +485,7 @@ firefoxNode:
     # Custom annotations for service
     annotations: {}
   # Size limit for DSH volume mounted in container (if not set, default is "1Gi")
-  dshmVolumeSizeLimit: 1Gi
+  dshmVolumeSizeLimit: 2Gi
```
diemol: Same comment as above.

charts/selenium-grid/values.yaml
```diff
@@ -398,6 +399,15 @@ chromeNode:
...
   #     claimName: my-pv-claim
 
+  # Keda scaled object configuration
+  autoscalingEnabled: false
+  maxReplicaCount: 8
+  hpa:
+    url: http://selenium-hub.default:4444/graphql # Repalce your http graphql url here
+    browserName: chrome
+    # browserVersion: '91.0' # Optional. Only required when supporting multiple versions of browser in your Selenium Grid.
+    unsafeSsl : 'true' # Optional
```
diemol, line 409: What does this mean?`unsafeSsl`?

charts/selenium-grid/values.yaml
```diff
@@ -487,17 +497,17 @@ firefoxNode:
     # failureThreshold: 120
     # periodSeconds: 5
   # Time to wait for pod termination
-  terminationGracePeriodSeconds: 30
+  terminationGracePeriodSeconds: 3600
```
diemol: Same comment as above.

charts/selenium-grid/values.yaml
```diff
@@ -487,17 +497,17 @@ firefoxNode:
...
-    #         while curl 127.0.0.1:5555/status; do sleep 1; done
+  lifecycle: 
+    preStop:
+      exec:
+        command:
+          - bash
+          - -c
+          - |
+            curl -X POST 127.0.0.1:5555/se/grid/node/drain --header 'X-REGISTRATION-SECRET;' && \
+            while curl 127.0.0.1:5555/status; do sleep 1; done;
```
diemol, line 510: Same comment as above.

diemol (changes requested):
Thanks for this PR, @prashanth-volvocars!

I left some comments.

charts/selenium-grid/values.yaml
prashanth-volvocars: If the autoscaler chooses to kill a pod that's running a test currently, we would have at most one hour for the running test to complete rather than 30 seconds. 
If the test doesn't complete within the configured grace period it would result in failure so giving it the maxium allowed time would help reduce these failures.

charts/selenium-grid/values.yaml
prashanth-volvocars: Makes sense. Will add the check

charts/selenium-grid/values.yaml
diemol: Wouldn't this affect the setup of users who are already having this configuration? Isn't there an alternative to this hard coded value? One of the reasons we added the draining after X sessions feature was to enable this use case, so the pod could live until the session is done.

msvticket: To expand on what @diemol writes: 

Wouldn't it be more straight forward to use ScaledJob combined with draining after one session?
I wrote about it here:

https://github.com/SeleniumHQ/selenium/issues/9845#issuecomment-1206257437

prashanth-volvocars: @diemol Killing after x sessions wouldn’t help. Consider this scenario where you create two sessions, one runs for 60 mins and another 1 min. As soon as the first is done, the hpa will try to bring down one of the pod irrespective of the configured x sessions. We have no control on which pod is chosen by hpa. This is not current flow but will happen when KEDA scales down.

Also I don’t understand the implication that would be caused by increasing it even for current users. No one would want their tests to be interrupted. Also with this value has no effect unless we have prestop lifecycle hook enabled. Even if that is enabled its gonna delay the pod termination only until the tes session is completed, which i think be desired behaviour

prashanth-volvocars: @msvticket Jobs doesn’t work great, if you are on severless kubernetes platform like EKS fargate. One new node is created for each pod and everytime it has to download the full browser node image and to create the node in itself combined takes around 90 seconds. Same would the case with GKE autopilot. Also i don’t see a point in killing the pod after just one session. Imagine a organisation of our scale, we need close to 75 pods running during the working hours. Imagine the number of sessions and how many times we would be creating  and killing pods.

The best part about KEDA scaler is that it works for both pods and jobs so if you think jobs would suit your use case better, its just matter of swapping the scaled object resource with scaled jobs. Let me know if you need any assistance, i would be ahppy to help

charts/selenium-grid/values.yaml
prashanth-volvocars: I had the same thought too, but created the PR in the excitement of sharing it😁. Will add this too.

charts/selenium-grid/values.yaml
prashanth-volvocars: We had some issues while running it along with video recording which vanished as soon as we increased the limit. Think it shouldn’t be part of this PR. Will revert this.

charts/selenium-grid/values.yaml
diemol: @prashanth-volvocars draining after X sessions will make the container exit and ideally the pod should stop, am I missing something?

charts/selenium-grid/values.yaml
```diff
@@ -621,6 +637,12 @@ edgeNode:
   # - name: my-extra-volume-from-pvc
   #   persistentVolumeClaim:
   #     claimName: my-pv-claim
+  # Keda scaled object configuration
+  autoscalingEnabled: false
+  maxReplicaCount: 8
+  hpa:
+    url: http://selenium-hub.default:4444/graphql # Repalce your http graphql url here
+    browserName: MicrosoftEdge
```
Wolfe1, line 645: If you was Edge to work you will need to have "sessionBrowserName" as well set to "msedge"

Docs: https://keda.sh/docs/2.8/scalers/selenium-grid-scaler/
Issue regarding this: https://github.com/kedacore/keda/issues/2709

charts/selenium-grid/values.yaml
Wolfe1: @diemol https://keda.sh/docs/2.8/scalers/selenium-grid-scaler/

"Skip certificate validation when connecting over HTTPS."

charts/selenium-grid/values.yaml
msvticket: I'd say that for most users ScaledJobs would be the best fit so this should be the default in the helm chart. Creating a pod is normally not expensive in Kubernetes; Fargate is a special case. But sure, you can skip that and I guess it's me or @diemol that will implement it.

prashanth-volvocars: @diemol yes, we no longer need X sessions as the pod can handle sessions as long as there are pending sessions in the queue.

prashanth-volvocars: @msvticket The default should be what works for every environment, not for a specific few. Jobs works for a few, not all, but pod works for all. If you think a scaled job suits your need best, it's available as an option that you can very well enable, but making it default will entirely break it for specific environments. Hope you understand.

msvticket: What option? I don't see any mention of it in the chart.

qalinn:
@prashanth-volvocars Any news regarding this PR?

QATew:
@prashanth-volvocars @diemol  Any updates on this?

prashanth-volvocars:
@diemol i have added/addressed all the concerns raised. Please take a look

diemol:
I will have a look after releasing 4.10.0, so in June.

jcputney:
@diemol any updates on this and #1854 ? It seems to add a lot of needed features to this PR

Issue #1754, opened 2022-12-22
How to make my Selenium-Hub URL password protected?


omdhimar28:
### What happened?

I have deployed my selenium grid architecture using docker images in my kubernetes cluster. Also I have made my selenium hub service publicly accessible via ingress. I want to make my publicy accessible URL password protected by some basic level authentication. For example, by changing config file for selenium hub or grid.

![image](https://user-images.githubusercontent.com/71244798/209128756-2f264494-af48-4241-be71-8dc1477d829f.png)

### Command used to start Selenium Grid with Docker

```shell
Selenium-hub deployment.yaml :

apiVersion: apps/v1
kind: Deployment
metadata:
  name: selenium-hub
  namespace: selenium
  labels:
    app: selenium-hub
spec:
  replicas: 1
  selector:
    matchLabels:
      app: selenium-hub
  template:
    metadata:
      labels:
        app: selenium-hub
    spec:
      containers:
      - name: selenium-hub
        image: selenium/hub:latest
        ports:
          - containerPort: 4444
          - containerPort: 4443
          - containerPort: 4442
```

### Relevant log output

```shell
How to make my Selenium URL password protected?

Reference URL:
https://www.selenium.dev/documentation/grid/configuration/toml_options/#basic-auth-enabled
```

### Operating System

Ubuntu

### Docker Selenium version (tag)

Hub version 4.6

Please help!!!
@diemol

omdhimar28:
Is there anyway that I can do some changes in the configuration file as shown in this link "https://www.selenium.dev/documentation/grid/configuration/toml_options/#basic-auth-enabled" and How can I do that using Dockerfile as I need to do change the configuration in docker image!!!

Please Help! @diemol

ghost:
@omdhimar28 I added the ability to use `config.toml` files on the hub.

You can check out this compose file if you are interested on how to add basicauth:

https://github.com/SeleniumHQ/docker-selenium/pull/1757/files#diff-a627c055a568a3c52aef1a4363aefcc87d8f5b0595ad5d1714693d902bc23142

Obviously, the PR needs to be approved and merged first before it is available to everyone.

omdhimar28:
Thanks @Xcalizorz , this will for sure help a lot of users.
Do you have any idea about how can I edit the config.toml file before running my docker image in a container.
Like if I can just do some kind of changes in my Dockerfile.

For example: [Dockerfile]

FROM selenium/hub:4.7.2-20221219
[...]

PR #1757, opened 2022-12-27
Allow adding `config.toml` to Selenium-Hub

ghost:

### Description

We now read the config file as well. The base Dockerfile creates the `config.toml` every time & leaves it empty. Hence, there will be no issues, if a user decides to not use the `config.toml`.

### Motivation and Context

It is not possible to load a personalized `config.toml` for the Selenium-Hub, hence adding things like [basic auth](https://www.selenium.dev/documentation/grid/configuration/toml_options/#basic-auth-enabled) is not easily possible from the outside (as asked for in #1754 ).

### Screenshots

locked

unlocked

This PR enables users to insert their own `config.toml` as they wish.

### Types of changes

- [X] New feature (non-breaking change which adds functionality)

### Checklist

- [X] I have read the [contributing](https://selenium.dev/documentation/en/contributing/) document.

### DIFF
--- a/Hub/example-config.toml
+++ b/Hub/example-config.toml
@@ -1,0 +1,3 @@
+[router]
+username = "admin"
+password = "myStrongPassword"
--- a/Hub/start-selenium-grid-hub.sh
+++ b/Hub/start-selenium-grid-hub.sh
@@ -36,6 +36,7 @@ java ${JAVA_OPTS:-$SE_JAVA_OPTS} -Dwebdriver.http.factory=jdk-http-client \
   --session-retry-interval ${SE_SESSION_RETRY_INTERVAL} \
   --relax-checks ${SE_RELAX_CHECKS} \
   --bind-host ${SE_BIND_HOST} \
+  --config /opt/selenium/config.toml \
   ${HOST_CONFIG} \
   ${PORT_CONFIG} \
   ${SE_OPTS}
--- a/docker-compose-v3-basicauth.yml
+++ b/docker-compose-v3-basicauth.yml
@@ -1,0 +1,44 @@
+# To execute this docker-compose yml file use `docker-compose -f docker-compose-v3.yml up`
+# Add the `-d` flag at the end for detached execution
+# To stop the execution, hit Ctrl+C, and then `docker-compose -f docker-compose-v3.yml down`
+version: "3"
+services:
+  chrome:
+    image: selenium/node-chrome:4.7.2-20221219
+    shm_size: 2gb
+    depends_on:
+      - selenium-hub
+    environment:
+      - SE_EVENT_BUS_HOST=selenium-hub
+      - SE_EVENT_BUS_PUBLISH_PORT=4442
+      - SE_EVENT_BUS_SUBSCRIBE_PORT=4443
+
+  edge:
+    image: selenium/node-edge:4.7.2-20221219
+    shm_size: 2gb
+    depends_on:
+      - selenium-hub
+    environment:
+      - SE_EVENT_BUS_HOST=selenium-hub
+      - SE_EVENT_BUS_PUBLISH_PORT=4442
+      - SE_EVENT_BUS_SUBSCRIBE_PORT=4443
+
+  firefox:
+    image: selenium/node-firefox:4.7.2-20221219
+    shm_size: 2gb
+    depends_on:
+      - selenium-hub
+    environment:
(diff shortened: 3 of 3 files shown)


jamesmortensen:
Hi @Xcalizorz thanks for working on this.  The screenshots demonstrate that it seems to be working.  

I have observed that all of the user-editable files (sample docker-compose.yml and other sample files), are generally kept in the root. I think one advantage is that images won't get accidentally built with the modified example file.  Because it's in the Hub folder, I believe the example-config.toml would still get copied into the image even if one weren't explicitly using the custom config. 

What do you think about creating a generate_config file to generate the config.toml, similar to what we have in the Standalone folder?  Before I saw your pull request, I was playing around with something like this as a derivative image:

**append-basic-auth:**
```
echo "[router]" >> "$FILENAME"
echo "username = \"$SEL_AUTH_USERNAME\"" >> "$FILENAME"
echo "password = \"$SEL_AUTH_PASSWORD\"" >> "$FILENAME"
```

**Dockerfile:**
```
FROM seleniarm/hub:latest

USER root

COPY append-basic-auth /opt/bin/
COPY start-selenium-grid-hub.sh /opt/bin/

RUN cat /opt/bin/append-basic-auth >> /opt/bin/generate_config && chmod 755 /opt/bin/generate_config

USER seluser
```

Then inside of start-selenium-grid-hub.sh I added:
```
FILENAME=/opt/selenium/config.toml /opt/bin/generate_config
```

and I then added the --config flag to the java command so the Hub would read it.

The only thing I didn't figure out was how to make the auth variables optional yet, since I was making a derivative image.  

Will any of this help?  What do you think?

ghost:
@jamesmortensen sounds interesting, I'll give it a try when I have time later this week.

If we want to ignore files in from the build, I could just add a simple `.dockerignore` file to the `./Hub` that ignores the example config file. The config.toml will be added to the image anyway (via `./Base/Dockerfile`) - the default is empty and can be changed by overriding via Volumes/Mounting.

What do you think about solving this via `.dockerignore`?

robinmatz:
Hey @Xcalizorz and @jamesmortensen ,

I have been following this conversation. I was wondering if the changes you are about to make will also support adding config other than basic auth, like relaying to an appium server (https://www.selenium.dev/documentation/grid/configuration/toml_options/#relaying-commands-to-a-service-endpoint-that-supports-webdriver), which would be the use case I happen to be interested in?

ghost:
@robinmatz you can add any config you want. My example config is using basic auth, but as long as you are able to add it into the `config.toml` you can do whatever you want.

Before this PR the Selenium hub would practically ignore the `config.toml`, now it's reading the config and acting upon it. As a user, you just have to override the default empty config as shown in the example here: https://github.com/SeleniumHQ/docker-selenium/pull/1757/files#diff-a627c055a568a3c52aef1a4363aefcc87d8f5b0595ad5d1714693d902bc23142R44

robinmatz:
Hey @Xcalizorz ,
thanks for clarifying! I might have misunderstood James' comment

> The only thing I didn't figure out was how to make the auth variables optional yet, since I was making a derivative image.

which made me unsure if allowing basic auth was the only objective of this PR.

jamesmortensen:
Hi @Xcalizorz  based on @robinmatz feedback and me doing a little more digging on my part, I can see that example-config.toml is not passed to the image during build time as I had mistakenly thought.  (I was confusing docker _context_ with the actual image being built).

I also validated that the grid works both with the config mapped via the volume property, with the username and password, and that it works without the config.toml and with no authentication and is backwards compatible.  I also confirmed that when the config.toml is included, the username and password is required in order to access the Grid.

Thus, I'll go ahead and merge these changes so that they'll be available in the next release.

ErcinDedeoglu:
Fantastic timing; I was investigating this issue :D 
When will it be released?

omdhimar28:
@jamesmortensen Looking forward for the release :)
When it will be released?

ErcinDedeoglu:
cc @diemol

jamesmortensen:
FYI: 4.8.0 was just released yesterday.  So the ability to add the config.toml to the Hub should be available now.    @omdhimar28 @ErcinDedeoglu :)

https://github.com/SeleniumHQ/docker-selenium/releases/tag/4.8.0-20230123

omdhimar28:
Thanks for the update @jamesmortensen :)

PR #1787, opened 2023-02-10
Using release-tag instead of commit-id on novnc and websockify and updated them

budtmo:
Changes:

- Using release tag instead of commit-id by downloading novnc and websockify  
- Updated them to the latest release version


diemol (approved):
Thank you, @budtmo! I wanted to do this long time ago! 🎉

Issue #1788, opened 2023-02-13
[🐛 Bug]: Language is hardcoded in generated wrapper scripts


adrianhj:
### What happened?

With recent changes to the Docker images our test-suites began failing due to incorrect formatting of dates within our application; we temporarily pinned back to the latest working version whilst investigating.

Reviewing recent changes, https://github.com/SeleniumHQ/docker-selenium/pull/1778 seems to be the culprit.

The changes to the `wrap_*_binary` scripts evaluate LANGUAGE outside of the generated script and do not escape variables within the heredoc causing it to hard-code the default LANGUAGE provided by the underlying image (en_US.UTF8) in the generated wrapper script.

### Command used to start Selenium Grid with Docker

```shell
Using `selenium-chrome` stand-alone image.

version: "3.7"
services:
  selenium-chrome:
    image: selenium/standalone-chrome:latest
    volumes:
      - /dev/shm:/dev/shm
    environment:
      - LANGUAGE=en_GB.UTF8
      - SCREEN_WIDTH=1920
      - SCREEN_HEIGHT=1080
      - SE_SESSION_REQUEST_TIMEOUT=500
      - SE_NODE_MAX_SESSIONS=2
      - SE_NODE_OVERRIDE_MAX_SESSIONS=true
      - SE_NODE_SESSION_TIMEOUT=500
    profiles:
      - chrome
```

### Relevant log output

```shell
$ docker run -it --rm selenium/standalone-chrome:110.0 cat /usr/bin/google-chrome

Unable to find image 'selenium/standalone-chrome:110.0' locally
110.0: Pulling from selenium/standalone-chrome
Digest: sha256:df7472a13463ff5e5e444bb485460ba1082c779469cbaa1f86b54997335761eb
Status: Downloaded newer image for selenium/standalone-chrome:110.0

#!/bin/bash

# umask 002 ensures default permissions of files are 664 (rw-rw-r--) and directories are 775 (rwxrwxr-x).
umask 002

# Set language environment variable
export LANGUAGE="en_US.UTF-8"

# Note: exec -a below is a bashism.
exec -a "$0" "/opt/google/chrome/google-chrome-base" --no-sandbox "$@"
```

### Operating System

Linux

### Docker Selenium version (tag)

110.0-20230210

jamesmortensen:
Fixed by #1789

PR #1778, opened 2023-01-24
Refine lang setting

jsa34:
**Thanks for contributing to the Docker-Selenium project!**
**A PR well described will help maintainers to quickly review and merge it**

Before submitting your PR, please check our [contributing](https://selenium.dev/documentation/en/contributing/) guidelines, applied for this repository.
Avoid large PRs, help reviewers by making them as simple and short as possible.

### Description

Sets the LANGUAGE env var as part of the binary wrapper script, instead of outside it,

### Motivation and Context

Resolves #1750.
Previous attempt didn't permeate to the node container in dynamic grid setup.

### Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### Checklist

- [x] I have read the [contributing](https://selenium.dev/documentation/en/contributing/) document.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

Issue #1750, opened 2022-12-19
[🐛 Bug]: Setting lang attribute doesn't apply to the test containers


jsa34:
### What happened?

When trying to run with a different locale, trying all possible ways to set this via the chromedriver doesn't seem to work.

Apparently there is an issue with setting this, and needs to be done at the env variable level.  This was previously identified:
https://github.com/elgalu/docker-selenium/issues/159

And I am still seeing the same behaviour in the current docker selenium images.

### Command used to start Selenium Grid with Docker

```shell
Using dynamic grid as documented on README.md
```

### Relevant log output

```shell
N/A

However, no matter what I set, the browser always is set to en_US and en as the languages in the settings section of the browser when inspecting.
```

### Operating System

Centos 9

### Docker Selenium version (tag)

4.5.0

jsa34:
Looks like this still doesn't work....

ChromeOptions set:
```
        ChromeOptions chromeOptions = getBaseChromeOptions();
        Map<String, Object> prefs = new HashMap<>();
        chromeOptions.addArguments("--no-sandbox");
        chromeOptions.addArguments("--verbose");
        chromeOptions.addArguments("--disable-dev-shm-usage");
        chromeOptions.addArguments("--safebrowsing-disable-download-protection");
        **chromeOptions.addArguments("--lang=en-GB");**
        chromeOptions.setCapability(CapabilityType.ACCEPT_INSECURE_CERTS, true);
        prefs.put("safebrowsing.enabled", "true");
        chromeOptions.setExperimentalOption("prefs", prefs);
        return chromeOptions;
```

And the lang attribute is send in the prefs when node container is created:
```
11:31:26.430 INFO [LocalDistributor.newSession] - Session request received by the Distributor:
 [Capabilities {acceptInsecureCerts: true, browserName: chrome, bstack:options: {buildName: Automation 24-01-2023, projectName: utomation}, goog:chromeOptions: {args: [--no-sandbox, --verbose, --disable-dev-shm-usage, --safebrowsing-disable-down..., --lang=en-GB], extensions: [], prefs: {safebrowsing.enabled: true}}, se:recordVideo: false, se:screenResolution: 1920x1080, se:timeZone: Europe/London}]
```

I wanted to check before I submit a new PR.  Should I set the env variable in the wrapper script instead of outside?  It looks like it's setting the LANGUAGE env variable but in the "default" node container, not the dynamically generated actual node containers for test runs?

[...]

Issue #1671, opened 2022-09-07
[🚀 Feature]: Are there parameters that support setting browser language?


dzjwwm:
### Feature and motivation

After the browser starts, the default is English. I want to set it to Chinese. So, I set it in the code, but it doesn't take effect

### Usage example

```
from selenium import webdriver
import time

chrome_options = webdriver.ChromeOptions()
chrome_options.add_argument('--lang=zh_CN.UTF-8')
driver = webdriver.Remote(command_executor='http://localhost:4444', options=chrome_options)
driver.get("http://192.168.16.100:81")
driver.maximize_window()
time.sleep(5)
driver.quit()

```

jamesmortensen:
Hi @tl31707chy I did some digging and found this issue:  https://github.com/SeleniumHQ/docker-selenium/issues/1009. Do you think the issue you're facing is the same here?  It seems like this may be a Chrome bug.

jamesmortensen:
Thanks to https://stackoverflow.com/a/41893197/552792, I can launch a standalone-chrome container, and then in bash, I can start chrome using the LANGUAGE environment variable:

```bash
$ LANGUAGE=zh_CN.UTF-8 google-chrome
```

Then I get this result: 
<img width="1435" alt="Screenshot 2022-09-07 at 7 24 55 PM" src="https://user-images.githubusercontent.com/1315816/188896669-3f710004-b10e-499b-a869-958b6294f81e.png">

Since changing the language via an environment variable works, what you can do is start the container and set the environment variable at container launch time:

```bash
$ docker run --rm -it -p 4444:4444 -p 7900:7900 --shm-size 2g -e LANGUAGE=zh_CN.UTF-8 selenium/standalone-chrome:latest
```

Additionally, not sure if the chrome_option specifying the language does anything or not.  You can try with and without to see if it's needed.

```python
from selenium import webdriver
import time

chrome_options = webdriver.ChromeOptions()
driver = webdriver.Remote(command_executor='http://localhost:4444', options=chrome_options)
driver.get("http://example.com")
driver.maximize_window()
time.sleep(5)
driver.quit()
```

dzjwwm:
> 感谢https://stackoverflow.com/a/41893197/552792，我可以启动一个独立 chrome 容器，然后在 bash 中，我可以使用 LANGUAGE 环境变量启动 chrome：
> 
> ```shell
> $ LANGUAGE=zh_CN.UTF-8 google-chrome
> ```
> 
[...]

PR #1789, opened 2023-02-13
fix: ensure language is not hardcoded in generated wrapper scripts

adrianhj:
### Description

Ensures LANGUAGE is not hardcoded in standalone images by moving logic into the generated wrapper script.

### Motivation and Context
See #1788 for related issue.

I have interpreted the original intention in #1778 to attempt to translate `--lang` to `LANGUAGE` at runtime, not build-time.

### Types of changes

### Checklist

### DIFF
--- a/NodeChrome/wrap_chrome_binary
+++ b/NodeChrome/wrap_chrome_binary
@@ -4,22 +4,22 @@ WRAPPER_PATH=$(readlink -f /usr/bin/google-chrome)
 BASE_PATH="$WRAPPER_PATH-base"
 mv "$WRAPPER_PATH" "$BASE_PATH"
 
-# Debian/Ubuntu seems to not respect --lang, it instead needs to be a LANGUAGE environment var
-# See: https://stackoverflow.com/a/41893197/359999
-for var in "$@"; do
-   if [[ $var == --lang=* ]]; then
-      LANGUAGE=${var//--lang=}
-   fi
-done
-
 cat > "$WRAPPER_PATH" <<_EOF
 #!/bin/bash
 
 # umask 002 ensures default permissions of files are 664 (rw-rw-r--) and directories are 775 (rwxrwxr-x).
 umask 002
 
+# Debian/Ubuntu seems to not respect --lang, it instead needs to be a LANGUAGE environment var
+# See: https://stackoverflow.com/a/41893197/359999
+for var in "\$@"; do
+   if [[ \$var == --lang=* ]]; then
+      LANGUAGE=\${var//--lang=}
+   fi
+done
+
 # Set language environment variable
-export LANGUAGE="$LANGUAGE"
+export LANGUAGE="\$LANGUAGE"
 
 # Note: exec -a below is a bashism.
 exec -a "\$0" "$BASE_PATH" --no-sandbox "\$@"
--- a/NodeEdge/wrap_edge_binary
+++ b/NodeEdge/wrap_edge_binary
@@ -4,22 +4,22 @@ WRAPPER_PATH=$(readlink -f /usr/bin/microsoft-edge)
 BASE_PATH="$WRAPPER_PATH-base"
 mv "$WRAPPER_PATH" "$BASE_PATH"
 
-# Debian/Ubuntu seems to not respect --lang, it instead needs to be a LANGUAGE environment var
-# See: https://stackoverflow.com/a/41893197/359999
-for var in "$@"; do
-   if [[ $var == --lang=* ]]; then
-      LANGUAGE=${var//--lang=}
-   fi
-done
-
 cat > "$WRAPPER_PATH" <<_EOF
 #!/bin/bash
 
 # umask 002 ensures default permissions of files are 664 (rw-rw-r--) and directories are 775 (rwxrwxr-x).
 umask 002
 
+# Debian/Ubuntu seems to not respect --lang, it instead needs to be a LANGUAGE environment var
+# See: https://stackoverflow.com/a/41893197/359999
+for var in "\$@"; do
+   if [[ \$var == --lang=* ]]; then
+      LANGUAGE=\${var//--lang=}
+   fi
+done
+
 # Set language environment variable
-export LANGUAGE="$LANGUAGE"
+export LANGUAGE="\$LANGUAGE"
 
 # Note: exec -a below is a bashism.
 exec -a "\$0" "$BASE_PATH" --no-sandbox "\$@"


adrianhj:
Output of the generated script after `make chrome`:

```shell
$ docker run -it --rm selenium/node-chrome:4.1.0-20230213 cat /usr/bin/google-chrome
#!/bin/bash

# umask 002 ensures default permissions of files are 664 (rw-rw-r--) and directories are 775 (rwxrwxr-x).
umask 002

# Debian/Ubuntu seems to not respect --lang, it instead needs to be a LANGUAGE environment var
# See: https://stackoverflow.com/a/41893197/359999
for var in "$@"; do
   if [[ $var == --lang=* ]]; then
      LANGUAGE=${var//--lang=}
   fi
done

# Set language environment variable
export LANGUAGE="$LANGUAGE"

# Note: exec -a below is a bashism.
exec -a "$0" "/opt/google/chrome/google-chrome-base" --no-sandbox "$@"
```

jamesmortensen:
Hi @adrianhj Thanks for the PR.  I'm attempting to review the changes, but I'm not sure how to validate them... Can you help me understand the difference between what we have now and what the changes do?  Here's some context:

Back in September, someone had asked how to change language at run-time:  https://github.com/SeleniumHQ/docker-selenium/issues/1671#issuecomment-1239436127

For example, I could do this to have the browser show text in Chinese:

```
docker run --rm -it -p 4444:4444 -p 7900:7900 --shm-size 2g -e LANGUAGE=zh_CN.UTF-8 selenium/standalone-chrome:latest
```

Would you please be able to help me understand what the script changes are doing and how I can validate the difference between the current revision and the changes, perhaps with an example with --lang?   Thank you! :)

adrianhj:
@jamesmortensen `LANGUAGE` is indeed the correct way as per [How to set your browser's locale > Linux](https://developer.chrome.com/docs/extensions/reference/i18n/#linux), and is also what we were relying on and what seems to have regressed.

This broke with [4.8.0-20230131](https://github.com/SeleniumHQ/docker-selenium/releases/tag/4.8.0-20230131) due to https://github.com/SeleniumHQ/docker-selenium/pull/1778 (original issue https://github.com/SeleniumHQ/docker-selenium/issues/1750).

There is not a huge amount of information on the original #1750 issue, but it seems that the outcome in #1778 was to provide some behaviour that attempts to adapt the `--lang` flag that is used by Windows/OSX to change locale to the LANGUAGE environment variable that is used by Linux (see other sections in earlier link).

Whether this is the right design choice or not for selenium-standalone I'm not sure, the original issue author and reviewer may be better off to elaborate here (https://github.com/SeleniumHQ/selenium/pull/11629 involving same pair seem to be opened since attempting another fix elsewhere with commentary suggesting the one in #1778 was unsuccesful).

In terms of the issue I raised in #1788, this can be validated by first running `4.8.0-20230123`, the version prior to #1778 being released, where this behaviour works fine using your example for `zh_CN`:

1. Launch `selenium/standalone-chrome:4.8.0-20230123` with LANGUAGE set to `zh_CN.UTF-8`

```shell
docker run --rm -it -p 4444:4444 -p 7900:7900 --shm-size 2g -e LANGUAGE=zh_CN.UTF-8 selenium/standalone-chrome:4.8.0-20230123
```

2. Accessing the image via noVNC on `:7900`, launch a shell and validate that the language is as expected via `env | grep -i language`:

3. From the same shell, launch via `google-chrome` and inspect the language of Google Chrome's UI + `navigator.language` in the JS console

Output from above steps:

Screenshot 2023-02-15 at 09 52 04

Now repeat the above but with `selenium/standalone-chrome:latest`, below now shows this having broken:

image

The issue stems from the wrapper generation scripts that generated the wrapped `google-chrome` entries having been incorrectly implemented. You can see this by issuing e.g. `cat $(which google-chrome)` to review the script that was generated:

* On 4.8.0-20230123:

```shell
$ seluser@43848d76539d:/$ cat $(which google-chrome)
#!/bin/bash

# umask 002 ensures default permissions of files are 664 (rw-rw-r--) and directories are 775 (rwxrwxr-x).
umask 002

# Note: exec -a below is a bashism.
exec -a "$0" "/opt/google/chrome/google-chrome-base" --no-sandbox "$@"
```

* On latest:

```shell
seluser@d89cc1f17477:/$ cat $(which google-chrome)
#!/bin/bash

# umask 002 ensures default permissions of files are 664 (rw-rw-r--) and directories are 775 (rwxrwxr-x).
umask 002

# Set language environment variable
export LANGUAGE="en_US.UTF-8"

# Note: exec -a below is a bashism.
exec -a "$0" "/opt/google/chrome/google-chrome-base" --no-sandbox "$@"
```

Language is now always hardcoded to en_US.UTF8 as an outcome of #1778.

In terms of validating this PR, after applying this PR and building via `make standalone_chrome`, the generated `google-chrome` wrapper script in the produced image is as per my earlier comment, and will now use LANGUAGE from its outer environment if no `--lang` is passed:

Screenshot 2023-02-15 at 10 08 50

Language can also be overridden via `--lang`, which I believe was the original intention of #1778:

Screenshot 2023-02-15 at 10 10 49

If there is any more information that would be of use please let me know.

jamesmortensen:
Thanks so much for the details. I'll try over the weekend or in the evening with the latest so I can reproduce.  (my "latest" was about 4 months old.)

jamesmortensen:
I validated that this is working using the following examples:

First, I built and started the container image:
```
git pull https://github.com/adrianhj/docker-selenium.git
cd docker-selenium && make standalone_chrome
docker run -it -p 4444:4444 -p 7900:7900 selenium/standalone-chrome:4.5.0-20230216
```

Then I ran a test Python script (using a shell script to create a virtualenv):

**run.sh**:
```
#!/usr/bin/env bash
# cd tests || true

if [ "${CI:-false}" = "false" ]; then
  pip3 install virtualenv | grep -v 'Requirement already satisfied'
  virtualenv selenium-webdriver-demo-python
  source selenium-webdriver-demo-python/bin/activate
fi

python -m pip install selenium==4.3.0 \
                      | grep -v 'Requirement already satisfied'

python remote-demo.py
ret_code=$?

if [ "${CI:-false}" = "false" ]; then
  deactivate
fi

exit $ret_code
```

**remote-demo.py**:
```
from selenium import webdriver
import time

chrome_options = webdriver.ChromeOptions()
#chrome_options.add_argument('--lang=zh_CN.UTF-8')
chrome_options.add_argument('--lang=es_US.UTF-8')
driver = webdriver.Remote(command_executor='http://localhost:4444', options=chrome_options)
driver.maximize_window()
driver.get("http://www.google.com")
time.sleep(5)
driver.quit()
```

I executed with both Chinese and Spanish, and both times I visually validated that the language changed in the browser.

Screenshot 2023-02-16 at 10 52 49 AM


Additionally, I also validated that the existing behavior, setting the language by default using the LANGUAGE environment variable, from September, still works as expected.  I used French as the default:

```
docker run -it -p 4444:4444 -p 7900:7900 -e LANGUAGE=fr_CA.UTF-8 selenium/standalone-chrome:20230216
```

Then I ran the same python code, but without setting the language in the code.  The language was still French.  

Screenshot 2023-02-16 at 10 50 47 AM


Additionally, if I set the language in the code as Chinese, then it overrides the default as expected.

Screenshot 2023-02-16 at 10 50 23 AM

jamesmortensen:
Thank you! @adrianhj

PR #1834, opened 2023-04-21
Custom Ingress Path for Helm Chart

JordanZimmitti:
### Description
Now that Selenium supports custom sub-paths for the UI, in order to take advantage of it in Kubernetes the helm chart needs the ingress `path` to be overwritable. I added a `path` key under the ingress section of the `values.yaml` file and made it backward compatible in the template so if no `path` key exists then it will use the default "/" path just like before.

### Motivation and Context
Deploy selenium with a custom domain sub-path using the helm chart

### Types of changes

### Checklist

### DIFF
--- a/Hub/start-selenium-grid-hub.sh
+++ b/Hub/start-selenium-grid-hub.sh
@@ -17,6 +17,11 @@ if [ ! -z "$SE_HUB_PORT" ]; then
   PORT_CONFIG="--port ${SE_HUB_PORT}"
 fi
 
+if [ ! -z "$SE_HUB_SUB_PATH" ]; then
+  echo "Using SE_HUB_SUB_PATH: ${SE_HUB_SUB_PATH}"
+  SUB_PATH_CONFIG="--sub-path ${SE_HUB_SUB_PATH}"
+fi
+
 EXTRA_LIBS="/opt/selenium/selenium-http-jdk-client.jar"
 
 if [ ! -z "$SE_ENABLE_TRACING" ]; then
@@ -39,4 +44,5 @@ java ${JAVA_OPTS:-$SE_JAVA_OPTS} -Dwebdriver.http.factory=jdk-http-client \
   --config /opt/selenium/config.toml \
   ${HOST_CONFIG} \
   ${PORT_CONFIG} \
+  ${SUB_PATH_CONFIG} \
   ${SE_OPTS}
--- a/Hub/start-selenium-grid-hub.sh
+++ b/Hub/start-selenium-grid-hub.sh
@@ -17,9 +17,9 @@ if [ ! -z "$SE_HUB_PORT" ]; then
   PORT_CONFIG="--port ${SE_HUB_PORT}"
 fi
 
-if [ ! -z "$SE_HUB_SUB_PATH" ]; then
-  echo "Using SE_HUB_SUB_PATH: ${SE_HUB_SUB_PATH}"
-  SUB_PATH_CONFIG="--sub-path ${SE_HUB_SUB_PATH}"
+if [ ! -z "$SE_SUB_PATH" ]; then
+  echo "Using SE_SUB_PATH: ${SE_SUB_PATH}"
+  SUB_PATH_CONFIG="--sub-path ${SE_SUB_PATH}"
 fi
 
 EXTRA_LIBS="/opt/selenium/selenium-http-jdk-client.jar"
--- a/charts/selenium-grid/Chart.yaml
+++ b/charts/selenium-grid/Chart.yaml
@@ -2,6 +2,6 @@ apiVersion: v2
 name: selenium-grid
 description: A Helm chart for creating a Selenium Grid Server in Kubernetes
 type: application
-version: 0.15.9
+version: 0.16.0
 appVersion: 4.9.0-20230421
 icon: https://github.com/SeleniumHQ/docker-selenium/raw/trunk/logo.png
--- a/charts/selenium-grid/Chart.yaml
+++ b/charts/selenium-grid/Chart.yaml
@@ -2,6 +2,6 @@ apiVersion: v2
 name: selenium-grid
 description: A Helm chart for creating a Selenium Grid Server in Kubernetes
 type: application
-version: 0.15.8
+version: 0.15.9
 appVersion: 4.9.0-20230421
 icon: https://github.com/SeleniumHQ/docker-selenium/raw/trunk/logo.png
--- a/charts/selenium-grid/templates/distributor-deployment.yaml
+++ b/charts/selenium-grid/templates/distributor-deployment.yaml
@@ -38,6 +38,8 @@ spec:
               value: {{ template "seleniumGrid.sessionQueue.fullname" . }}
             - name: SE_SESSION_QUEUE_PORT
               value: {{ .Values.components.sessionQueue.port | quote }}
+            - name: SE_SUB_PATH
+              value: {{ .Values.components.subPath }}
           {{- with .Values.components.extraEnvironmentVariables }}
             {{- tpl (toYaml .) $ | nindent 12 }}
           {{- end }}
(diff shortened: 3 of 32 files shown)


charts/selenium-grid/Chart.yaml
```diff
@@ -2,6 +2,6 @@ apiVersion: v2
 name: selenium-grid
 description: A Helm chart for creating a Selenium Grid Server in Kubernetes
 type: application
-version: 0.15.8
+version: 0.15.9
```
diemol: I would move this to 0.16.0 since there is a new feature.

diemol (commented):
I do not see how the sub path is getting configured in the Grid itself. Am I missing something?

JordanZimmitti:
@diemol 
> I do not see how the sub path is getting configured in the Grid itself. Am I missing something?

I was able to configure that part using the existing helm chart already by adding these lines here:

```yaml
hub:
    extraEnvironmentVariables:
        - name: SE_OPTS
           value: '--sub-path /<custom-sub-path>/'
  
```

which worked fine, but then could not update the ingress path itself since it was hard coded

charts/selenium-grid/Chart.yaml
JordanZimmitti: Makes sense, I updated the value to 0.16.0

diemol:
@JordanZimmitti would if be better if we add a new environment variable just to set the sub-path? If so, I can have a look at it unless you want to do it.

JordanZimmitti:
@diemol it did take me a little while to figure out how to set the subpath for the hub so having an env called `SUB_PATH` or something similar might be easier for people to interpret. I'll take a stab at it I think I might know where that needs to be updated

JordanZimmitti:
@diemol I updated the selenium hub start script to take a new environment variable called `SE_HUB_SUB_PATH` which will set the sub-path. Tested this locally with and without using the env variable and it worked as expected. I followed the naming convention that I saw being used in the script, if you would like something different let me know and I'll update it

Hub/start-selenium-grid-hub.sh
```diff
@@ -17,6 +17,11 @@ if [ ! -z "$SE_HUB_PORT" ]; then
   PORT_CONFIG="--port ${SE_HUB_PORT}"
 fi
 
+if [ ! -z "$SE_HUB_SUB_PATH" ]; then
```
diemol: I would just use `SE_SUB_PATH`. Why?

Because we should also use it in the [Standalone](https://github.com/SeleniumHQ/docker-selenium/blob/trunk/Standalone/start-selenium-standalone.sh) and the [Router](https://github.com/SeleniumHQ/docker-selenium/blob/trunk/Router/start-selenium-grid-router.sh).

diemol (changes requested):
Looks good, we just need to add it to a couple of places more. I'd be nice if you have time to put a couple of lines describing it on the main README. Thanks!

Hub/start-selenium-grid-hub.sh
JordanZimmitti: Sounds good will do!

JordanZimmitti:
@diemol Thanks for the suggestions! I updated the environment variable name to `$SE_SUB_PATH` as suggested and added it to the `Standalone` and `Router` scripts. I also added some documentation for using the environment variable in the `readme.md` in a spot that I think makes sense but if you want me to put it somewhere else let me know

diemol (commented):
Looks great!

Not sure if I am missing something, but shouldn't the `SE_SUB_PATH` variable be used somewhere in the chart?

JordanZimmitti:
@diemol 
> Looks great!
> 
> Not sure if I am missing something, but shouldn't the `SE_SUB_PATH` variable be used somewhere in the chart?

Since the docker file takes `SE_SUB_PATH` in as an environment variable it would work fine with the helm chart as is like this

```yaml
hub:
    extraEnvironmentVariables:
        - name: SE_SUB_PATH
           value: "/<custom-sub-path>/"
```

But I could break it out of `extraEnvironmentVariables` and make it its own key under the `hub` section if you want.
If we go down that path then I probably need to make a `sub_path` key under the `router` section too since we added it to the start script there as well correct?

diemol:
@JordanZimmitti sorry for the late reply, we had a long weekend here. 

>But I could break it out of extraEnvironmentVariables and make it its own key under the hub section if you want.
If we go down that path then I probably need to make a sub_path key under the router section too since we added it to the start script there as well correct?

I think ☝️ is the preferred solution so all places where the env var can be used get the option to be configured. Thanks!

JordanZimmitti:
@diemol Sounds good ill implement the changes in the next day or so!

JordanZimmitti:
@diemol 
I broke out the environment variable in the helm charts in the needed locations, let me know if you have any other suggestions!

charts/selenium-grid/templates/distributor-deployment.yaml
```diff
@@ -38,6 +38,8 @@ spec:
               value: {{ template "seleniumGrid.sessionQueue.fullname" . }}
             - name: SE_SESSION_QUEUE_PORT
               value: {{ .Values.components.sessionQueue.port | quote }}
+            - name: SE_SUB_PATH
+              value: {{ .Values.components.subPath }}
```
diemol: This is not needed here. The sub-path only applies to Router and Hub.

charts/selenium-grid/templates/event-bus-deployment.yaml
```diff
@@ -36,9 +36,12 @@ spec:
...
               protocol: TCP
-        {{- with .Values.components.extraEnvironmentVariables }}
-          env: {{- tpl (toYaml .) $ | nindent 12 }}
-        {{- end }}
+          env:
+            - name: SE_SUB_PATH
+              value: {{ .Values.components.subPath }}
+          {{- with .Values.components.extraEnvironmentVariables }}
+            {{- tpl (toYaml .) $ | nindent 12 }}
+          {{- end }}
```
diemol: This is not needed here. The sub-path only applies to Router and Hub.

charts/selenium-grid/templates/session-map-deployment.yaml
```diff
@@ -29,9 +29,12 @@ spec:
...
           imagePullPolicy: {{ .Values.components.sessionMap.imagePullPolicy }}
-        {{- with .Values.components.extraEnvironmentVariables }}
-          env: {{- tpl (toYaml .) $ | nindent 12 }}
-        {{- end }}
+          env:
+            - name: SE_SUB_PATH
+              value: {{ .Values.components.subPath }}
+          {{- with .Values.components.extraEnvironmentVariables }}
+            {{- tpl (toYaml .) $ | nindent 12 }}
+          {{- end }}
```
diemol: This is not needed here. The sub-path only applies to Router and Hub.

charts/selenium-grid/templates/session-queuer-deployment.yaml
```diff
@@ -29,9 +29,12 @@ spec:
...
           imagePullPolicy: {{ .Values.components.sessionQueue.imagePullPolicy }}
-        {{- with .Values.components.extraEnvironmentVariables }}
-          env: {{- tpl (toYaml .) $ | nindent 12 }}
-        {{- end }}
+          env:
+            - name: SE_SUB_PATH
+              value: {{ .Values.components.subPath }}
+          {{- with .Values.components.extraEnvironmentVariables }}
+            {{- tpl (toYaml .) $ | nindent 12 }}
+          {{- end }}
```
diemol: This is not needed here. The sub-path only applies to Router and Hub.

diemol (changes requested):
Thank you!

I just noted that the sub-path environment variable is being used in a few places we do not need it.

JordanZimmitti:
@diemol I removed the sub_path from components that don't need it based on your recommendation!

diemol (approved):
Thank you, @JordanZimmitti!

PR #2058, opened 2023-12-11
Guard against high `ulimit -n` in when starting vnc

### LINKED ISSUE #2045: [🐛 Bug]: Cannot connect through NoVNC
### What happened?

When upgrading `standalone-chrome` from `4.15.0-20231108` to `4.15.0-20231110`, the NoVNC web interface is not able to connect. Issue exists on the latest version `4.15.0-20231129` as well, I just tested in which version it started.

It's perpetually stuck in the "Connecting..." screen, the websocket being openend is not recieving any data.
image

The only difference between these two versions is the upgrade from Focal to Jammy in PR #1923

### Command used to start Selenium Grid with Docker (or Kubernetes)

```shell
version: "3"

services:
  selenium:
    image: selenium/standalone-chrome:4.15.0-20231110
    environment:
      - SE_VNC_NO_PASSWORD=1
    shm_size: 2gb
    ports:
      - ${EXPOSED_VNC_PORT:-7900}:7900
```

### Relevant log output

```shell
None that I can see
```

### Operating System

Arch Linux

### Docker Selenium version (tag or chart version)

4.15.0-20231110

Earlopain:
### Description
See #2045. Basically what https://github.com/SeleniumHQ/docker-selenium/issues/2045#issuecomment-1849497065 proposed. If the ulimit is high, defensively lower it.

### Motivation and Context
Recent versions of docker in combination with the upstream systemd unit files pass an incredibly high `ulimit -n` to the docker container, up to 1 billion. That causes minute high delays and CPU spinning when connecting to VNC while it enumerates all the file descriptors.

### Types of changes

### Checklist

### DIFF
--- a/NodeBase/start-vnc.sh
+++ b/NodeBase/start-vnc.sh
@@ -44,6 +44,17 @@ if [ "${START_XVFB:-$SE_START_XVFB}" = true ] ; then
       echo "Waiting for Xvfb..."
     done
 
+    # Guard against unreasonably high nofile limits. See https://github.com/SeleniumHQ/docker-selenium/issues/2045
+    if [[ $(ulimit -n) -gt 200000 ]]; then
+      echo "Trying to lower the open file descriptor limit from $(ulimit -n) to 65536."
+      ulimit -n 65536
+      if [ $? -eq 0 ]; then
+        echo "Successfully lowered the open file descriptor limit."
+      else
+        echo "Failed to lower the open file descriptor limit. This can result in delays when connecting to VNC."
+      fi
+    fi
+
     x11vnc ${X11VNC_OPTS} -forever -shared -rfbport ${VNC_PORT:-$SE_VNC_PORT} -rfbportv6 ${VNC_PORT:-$SE_VNC_PORT} -display ${DISPLAY}
   else
     echo "VNC won't start because SE_START_VNC is false."
--- a/NodeBase/start-vnc.sh
+++ b/NodeBase/start-vnc.sh
@@ -51,7 +51,7 @@ if [ "${START_XVFB:-$SE_START_XVFB}" = true ] ; then
       if [ $? -eq 0 ]; then
         echo "Successfully lowered the open file descriptor limit."
       else
-        echo "Failed to lower the open file descriptor limit. This can result in delays when connecting to VNC."
+        echo "The open file descriptor limit could not be updated. This can result in delays when connecting to VNC."
       fi
     fi
 
--- a/NodeBase/start-vnc.sh
+++ b/NodeBase/start-vnc.sh
@@ -45,13 +45,13 @@ if [ "${START_XVFB:-$SE_START_XVFB}" = true ] ; then
     done
 
     # Guard against unreasonably high nofile limits. See https://github.com/SeleniumHQ/docker-selenium/issues/2045
-    if [[ $(ulimit -n) -gt 200000 ]]; then
-      echo "Trying to lower the open file descriptor limit from $(ulimit -n) to 65536."
-      ulimit -n 65536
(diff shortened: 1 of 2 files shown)


NodeBase/start-vnc.sh
```diff
@@ -44,6 +44,17 @@ if [ "${START_XVFB:-$SE_START_XVFB}" = true ] ; then
...
+    if [[ $(ulimit -n) -gt 200000 ]]; then
+      echo "Trying to lower open file descriptor limit from $(ulimit -n) to 65536."
+      ulimit -n 65536
+      if [ $? -eq 0 ]; then
+        echo "Successfully lowered the open descriptor limit."
+      else
+        echo "Failed to lower the open file descriptor limit. This can result in delays when connecting to VNC."
+      fi
+    fi
+
```
diemol: What happens if the command fails? Does execution break or it just resumes and echoes the 2nd message?

Earlopain: It just carries on, this file doesn't have stop on error enabled. A few lines above is a similar construct.

I'm not even 100% sure that this command can even fail. Lowering the limit seems like something that always works but I'm not knowledgeable enough to say for certain so I added a bit of logging.

VietND96: I have tried a negative test by update wrong command, 2nd message echoes and next step can go
```bash
Trying to lower open file descriptor limit from 1048576 to 65536.
/opt/bin/start-vnc.sh: line 50: ulimit: -1: invalid option
ulimit: usage: ulimit [-SHabcdefiklmnpqrstuvxPT] [limit]
Failed to lower the open file descriptor limit. This can result in delays when connecting to VNC.
11/12/2023 16:56:48 passing arg to libvncserver: -rfbport
11/12/2023 16:56:48 passing arg to libvncserver: 5900
11/12/2023 16:56:48 passing arg to libvncserver: -rfbportv6
11/12/2023 16:56:48 passing arg to libvncserver: 5900
11/12/2023 16:56:48 -usepw: found /home/seluser/.vnc/passwd

```

diemol: If we are not sure if the command can fail, wouldn't it be a misleading message?

Earlopain: The file descriptor limit has been found to be high and lowering it failed, for whatever reason. That's exactly what the message is warning you about, if say your limit is set to 1 billion for one reason or another. Continuing on the user should be aware that things might not work like they expect them to.

The output above from vietnd96 introduced an intentional invalid command and it'll never look like that with the actual code.

VietND96: My test only verify status != 0 not break the code. To avoid misleading, I think we can change another message without word `Failed` something like
`echo "The open file descriptor limit could not be updated. This can result in delays when connecting to VNC."`

VietND96: I think the limit value, we can get from ENV or default, it helps user is able to set another value e.g `-e SE_VNC_ULIMIT=100000`
something like `ulimit -n ${SE_VNC_ULIMIT:-65536}`

Earlopain: Sure, done. Your version sounds better anyways

Earlopain: > I think the limit value, we can get from ENV or default, it helps user is able to set another value e.g `-e SE_VNC_ULIMIT=100000` something like `ulimit -n ${SE_VNC_ULIMIT:-65536}`

I pushed a commit, check it out. I dropped the notice about VNC connection delays since that isn't always applicable anymore.

diemol (approved):

VietND96:
Thank you, @Earlopain!

PR #3229, opened 2026-09-05
[ci] Build images once per PR and promote by tag instead of rebuilding

VietND96:
## The problem, measured

Every job in the PR suite builds the images it needs, from scratch.

| Workflow | Matrix jobs |
| --- | --- |
| `docker-test.yml` | 18 |
| `k8s-scaling-test.yml` | 9 |
| `helm-chart-test.yml` | 8 |
| `k8s-dynamic-grid-test.yml` | 2 |
| **Total** | **37** |

It is not only the four entries with `build-all: true`. Every test target names image targets as Make **prerequisites** — `test_video: video hub chrome firefox edge chromium` — so all 37 build something. Recent PR runs took **75–88 minutes**.

`deploy.yml` then built them a 38th time, so the artefact released was a *rebuild of the tested source* rather than the tested artefact.

## The design

Build once, publish to GHCR, have everything else pull.

| Tag | Meaning |
| --- | --- |
| `ghcr.io/<owner>/<image>:pr-<N>` | One set per pull request |
| `ghcr.io/<owner>/<image>:trunk-<sha>` | One set per trunk run |
| `ghcr.io/<owner>/<image>:main` | The last trunk set **whose tests passed** |

`build-images.yml` decides what each run actually needs:

| Mode | When | Result |
| --- | --- | --- |
| `reuse-main` | Nothing image-affecting changed | **Nothing is built at all** |
| `build-and-push` | Image content changed | Built once, multi-arch; 37 jobs pull |
| `build-in-job` | Fork PR | Exactly today's behaviour |

**Image-affecting** means the 21 directories holding a `Dockerfile`, or the `Makefile` (build args and pinned versions). That rule is deliberately coarse — an unnecessary build costs minutes, a missed one ships an untested image. Measured against the last 18 merges: **12 would rebuild, 6 would reuse `:main`.**

## Fork safety

Forks are detected by comparing head repository to this one, and **never receive credentials**. `pull_request_target` is deliberately not used — it would hand a registry write token to untrusted code. 3 of the last 60 PRs came from forks; they keep building in-job, unchanged.

## How tests consume the images

Test jobs pull and retag to the local names the compose files already expect, then set `SKIP_BUILD=true`. That turns every image-build target into a no-op via **one guarded block** at the end of the Makefile.

The 48 build recipes and 59 test targets are **untouched**. The block sits last because GNU Make lets the last definition of a target win — I had it at the top first and `SKIP_BUILD` silently did nothing, which is exactly the sort of thing that would have looked green and scanned nothing.

`gen_certs`, `prepare_resources` and `update_go` are deliberately **not** skipped: they produce working-tree files the tests read, not images.

```
make -n test_video                    -> 8 docker buildx build lines, 0 warnings
make -n test_video SKIP_BUILD=true    -> 0 docker buildx build lines
make -n build                         -> 30 build lines, unchanged
```

## Promotion, not rebuilding

`:main` is moved **only after the full suite passes**, by retagging the run-scoped `trunk-<sha>` manifest — so `:main` always denotes a tested tree and the digest never changes.

`deploy.yml` pulls `:main` and promotes it, falling back to building when `:main` does not exist — which it will not, on the first release after this lands.

## Cleanup

`pr-*` tags are deleted when the PR closes, plus a weekly sweep. The script only removes tags matching `pr-<digits>`, and only when **every** tag on that manifest matches, so a shared or release tag cannot be caught by it. Its image list is asserted to match the Makefile's `CI_IMAGES`.

## What is verified, and what is not

Verified locally:

- `SKIP_BUILD` contract, all three cases above.
- All five decision branches of the `decide` job, run as shell against real commits: fork, trunk push, forced, docs-only commit (`reuse-main`), `NodeChrome/` commit (`build-and-push`).
- All 27 workflow files parse; `mbake validate` passes; the two image lists match exactly.
- Build and pull steps are mutually exclusive in all four test workflows.

**Not verified, and it needs a real run:** none of this has executed in Actions. The first PR after merge is the real test. Two things I would watch specifically:

1. **The first run cannot use `:main`** — it does not exist yet, so the `decide` job will choose `build-and-push`. That is handled, but it means the speedup only shows from the second PR onward.
2. **GHCR package visibility.** The packages must allow the repository's `GITHUB_TOKEN` to pull. If `:pr-N` pushes succeed but test jobs fail to pull, that is the cause, and it is a package settings change rather than a workflow one.

I would merge this when you can watch a PR run through it, rather than at the end of a day.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01EGLHB3pjY3mGZBu4TyUsrm

### DIFF
--- a/.github/workflows/build-images.yml
+++ b/.github/workflows/build-images.yml
@@ -240,7 +240,7 @@ jobs:
 
       - name: Log in to GHCR
         if: inputs.push
-        run: echo "${{ secrets.GITHUB_TOKEN }}" | docker login ghcr.io -u "${{ github.actor }}" --password-stdin
+        run: echo "${{ secrets.SELENIUM_CI_TOKEN || secrets.GITHUB_TOKEN }}" | docker login ghcr.io -u "${{ secrets.SELENIUM_CI_USERNAME || github.actor }}" --password-stdin
 
       - name: Build images
         uses: nick-invision/retry@master
@@ -292,7 +292,7 @@ jobs:
         with:
           persist-credentials: false
       - name: Log in to GHCR
-        run: echo "${{ secrets.GITHUB_TOKEN }}" | docker login ghcr.io -u "${{ github.actor }}" --password-stdin
+        run: echo "${{ secrets.SELENIUM_CI_TOKEN || secrets.GITHUB_TOKEN }}" | docker login ghcr.io -u "${{ secrets.SELENIUM_CI_USERNAME || github.actor }}" --password-stdin
       - name: Merge the architecture tags
         run: |
           CI_REGISTRY="${{ needs.decide.outputs.registry }}" CI_TAG="${{ needs.decide.outputs.ci-tag }}" \
--- a/.github/workflows/build-images.yml
+++ b/.github/workflows/build-images.yml
@@ -1,0 +1,215 @@
+name: Build images once
+
+# Builds the image set at most once per run and publishes it to GHCR, so the ~37
+# test matrix jobs can pull instead of each building the images it needs. Every
+# test target names image targets as Make prerequisites, so before this they all
+# built something; recent pull request runs took 75-88 minutes.
+#
+# Three modes, decided by the `decide` job and honoured by every test workflow:
+#
+#   reuse-main     the change cannot affect image content, so nothing is built
+#                  and the tests run against the last :main
+#   build-and-push build once, push :pr-<N>, every test job pulls it
+#   build-in-job   fork pull requests, which have no registry credentials, keep
+#                  today's behaviour and build in each job
+#
+# Fork detection compares the head repository against this one. A fork run never
+# receives credentials and never pushes. pull_request_target is deliberately not
+# used: it would run untrusted code with a registry write token.
+
+on:
+  workflow_call:
+    inputs:
+      ci-tag:
+        description: 'Tag to publish. Defaults to pr-<number> or main.'
+        required: false
+        type: string
+        default: ''
+      force-build:
+        description: 'Build even when the change looks image-neutral'
+        required: false
+        type: boolean
+        default: false
+      push:
+        description: 'Push the built images. False builds only, for verification.'
+        required: false
+        type: boolean
+        default: true
+    outputs:
+      mode:
+        description: 'reuse-main, build-and-push, or build-in-job'
+        value: ${{ jobs.decide.outputs.mode }}
+      ci-tag:
+        description: 'Tag the test jobs should pull'
+        value: ${{ jobs.decide.outputs.ci-tag }}
+      registry:
+        description: 'Registry holding that tag'
+        value: ${{ jobs.decide.outputs.registry }}
+
+env:
+  CI_REGISTRY: ghcr.io/${{ github.repository_owner }}
+
+jobs:
+  decide:
+    name: Decide whether to build
+    runs-on: ubuntu-24.04
+    permissions:
+      contents: read
+      packages: read
+    outputs:
+      mode: ${{ steps.decide.outputs.mode }}
+      ci-tag: ${{ steps.decide.outputs.ci-tag }}
+      registry: ${{ steps.decide.outputs.registry }}
+    steps:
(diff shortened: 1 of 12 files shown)


VietND96:
## Review: do not merge

I reviewed this branch adversarially and found four blockers, two of which would ship broken releases. Recording them here and moving the PR back to draft.

### 1. Releases would be single-architecture

`deploy.yml` promotes `:main` by running `pull_ci_images`, which does `docker pull --platform linux/<arch>` and retags locally. `make release` then `docker push`es from that local store — so every published tag (`<version>-<date>`, major, minor, `latest`, browser tags) becomes **amd64-only**. arm64 users get `no matching manifest`.

Today's release builds `PLATFORMS=linux/amd64,linux/arm64` and pushes the index directly, so this is a regression introduced by the promote path. `merge`/`promote`/`alias` already do it correctly with `docker buildx imagetools create`, which is registry-side and preserves the index; deploy must copy `:main` → Docker Hub the same way.

### 2. `:main` can be tagged onto images no test ran against

`promote-main` gates on `!failure()`. The three test jobs gate on `contains(toJson(github.event.commits), '[skip test]') == false`, and a **skipped** job is not a failure — so `promote-main` runs with all three skipped.

This is not hypothetical: `deploy.yml` commits `"[skip test]"` itself after every release, and `update_tag_in_docs_and_files.sh` rewrites files under `Base/`, `NodeChrome/`, `Video/` and the `Makefile` — all hashed paths. So each release produces a new `src-<hash>`, skips every test, promotes it to `:main`, and the next release promotes that to Docker Hub. Must gate on `needs.<job>.result == 'success'`.

### 3. A partial `merge` wedges CI permanently

`decide` probes only `base:<tag>`, and `base` is the first entry in `CI_IMAGES`. If `merge_ci_images` fails or is cancelled partway (`cancel-in-progress: true` makes that routine), `base:src-<hash>` exists while later images do not. Every subsequent run then chooses `reuse` and hard-fails in `pull_ci_images`.

There is no recovery: `build-images.yml` has no `workflow_dispatch`, and `build-test.yml` never passes `force-build`. Needs a completeness marker written only after `merge` succeeds, plus a force-build escape hatch.

### 4. Nightly would test stable-core images

`build-images.yml` has a `core` input, but `build-test.yml` never forwards it. `nightly.yml` calls with event `schedule`, so `decide` resolves **stable** — while nightly's own deploy job still publishes **nightly**-core images. Nightly would stop testing what it ships. The test workflows also re-resolve the core themselves and get nightly, so `BASE_VERSION` in the test environment disagrees with the images being tested.

### Also found

- Per-architecture tags (`src-<hash>-amd64`) match none of the three cleanup modes' regexes, so two thirds of every build is retained forever — the exact problem the script exists to solve.
- A partial `alias` leaves images unaliased; the weekly orphan sweep then deletes manifests an open PR still needs, wedging that PR.
- `update_go` mutates hashed sources inside the build job, so `src-<hash>` is not genuinely content-addressed.
- `pull_ci_images`' architecture check treats a single (non-index) manifest as "wrong arch" and skips it silently.
- The cleanup workflow uses `GITHUB_TOKEN` for deletes while images are pushed with `SELENIUM_CI_TOKEN`; deletes will likely 403.
- The cleanup tests re-implement the deletion rules rather than calling them, so a bug in the real loop cannot be caught — poor for a script that deletes from a production registry.

### Scope

This bundles a CI redesign, registry-lifecycle tooling, a production-registry deletion script, and a KEDA scaler test-semantics change. The KEDA change switches a failing configuration to a passing one and removes the last coverage of KEDA's built-in scaler; it deserves its own PR and its own argument.

### What stands up

The `SKIP_BUILD` guard is correct and leaves `make build` untouched. Fork handling is right and avoids `pull_request_target`. The native per-architecture build is the correct answer to QEMU. `deploy.yml`'s hash check does fall back to a rebuild when the core or content disagrees. The K8s suite passed end to end against pulled images, which proves the core mechanism works.

The design is sound; the edges are not. Splitting it and fixing the four blockers is the way forward.

